PR #10, opened 2016-05-25
Estimator

superjax:
This merge represents a pretty significant change - I am now representing our state in urad and urad/s, rather than mrad and mrad/s.  This is because I noticed that very small gyro measurements were being truncated off, and the only way I could fix it was cranking up the attitude resolution.  I also changed the _accel_scale factor to scale up so that accelerometer readings are represented in um/s^s.  This again, is because I noticed that very small accelerometer measurements were being truncated off.  I did the math on the _accel_scale and _gyro_scale, and manually wrote in the actual scale factor.  Before, our accelerometer scale factor was 2.  Now it is something like 2394.  I know we are probably not getting any actual accuracy, but 2 seemed like a really dumb scale factor, especially since it makes the math easier later, and represents zero extra clock cycles.

I was also reading through the cleanflight estimator and I noticed that they reject any accel measurements greater than 1.15 G and less than 0.85 G.  I figured that was a good idea, so I implemented it.


src/sensors.c
```diff
@@ -29,16 +29,21 @@ void init_sensors(void)
   uint16_t acc1G;
   float gyro_scale;
   mpu6050_init(true, &acc1G, &gyro_scale);
-  _accel_scale = 9807 / acc1G; // convert to mm/s^2
-  _gyro_scale = (int32_t) (1e6 * gyro_scale); // convert to urad/s
+
+  _accel_scale = 2394; // convert to um/s^2
```
dpkoch: Works, but I don't love the magic numbers here. Can we get them from his acc1G and gyro_scale returns?

dpkoch:
Looks good! Aside from my comment above, I'm good with merging. One question though, were we going to get rid of psi?

src/sensors.c
superjax: So, the problem is that acc1G and gyro_scale in his code are not dynamic, and they are wrong.  I'm hopefully pull requesting some changes to his code so that they will be right.

superjax:
So, one of the reasons I did this big change is that the estimate of psi was previously horrible.  Now that I've made these changes, psi is actually really good.  I think that we might want to keep it around for now, and I'm considering adding estimation based on the magnetometer, for those who might want that.

src/sensors.c
superjax: I got the magic numbers figured out, and I have them based on Simon's numbers.  It took me a little bit to figure out what was going on.  I just committed a new version with his numbers

superjax:
I fixed the magic numbers that we were talking about earlier, and I fixed some weird build errors that I was having.  This should be ready to merge in as master.  Not really sure why it wasn't working before, I thought I had tested it.

PR #14, opened 2016-05-31
Updated mavlink submodule to point to commit with updated offboard control message

dpkoch:
So instead of adding bitfield for which fields are valid, I added one for which fields should be ignored. My thought was that ignoring is more of the exception, so you should only have to initialize the bitfield to something other than zero if you want that special case. Then you could also just test the value of the whole bitfield to quickly see if anything is being ignored, if you so desire. It's just a thought I had and not something I'm set on, if you don't like it I'm happy to do it the other way.

The relevant message definition is at https://github.com/BYU-MAGICC/mavlink/blob/master/message_definitions/v1.0/rosflight.xml#L38.

### DIFF
--- a/include/mavlink_receive.h
+++ b/include/mavlink_receive.h
@@ -1,3 +1,12 @@
 #pragma once
 
+#include <stdint.h>
+
+#include "mavlink.h"
+
+// global variable declarations
+extern mavlink_offboard_control_t _offboard_control;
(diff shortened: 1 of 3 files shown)


superjax:
I like that idea.  On the ROSflight side, I want to be able to make sure that each channel is active after muxing.  So I think the active member is better there, however for the offboard control I think this makes sense.

PR #15, opened 2016-06-02
Mavlink refactor

dpkoch:
Changes:
* Refactored mavlink_stream to reduce code duplication and make it easier to add new streams
* Fixed bug in handling mavlink param_request_read messages
* Added optional mavlink stream for servo outputs. This stream is disabled by default, so you'll need to set the STRM_SERVO parameter to some rate in Hz to turn it on


superjax:
Cool!  To be honest, I'm not 100% sure how all the mavlink stuff works.  I'll spend some time figuring it out one of these days, I should understand better how it works, but I like the changes I see here!

PR #20, opened 2016-06-16
Added functionality for sending named debug values

dpkoch:


superjax:
I can't get fcu_io to connect with the naze when I'm passing the named value stuff across.  Can you take a look at the most recent commit on this branch, and tell me if I'm doing something wrong?

dpkoch:
I did test it before pushing and was able to send named value messages at 10Hz no problem, faster probably would've worked too, I just didn't try.

My best guess is that your fcu_io2 version is out of sync. Did you
 1. Checkout the `mavlink_debug` branch?
 2. After checking it out run `git submodule update` to bring the mavlink library version up to date? The submodule commit on the master branch has old message definitions.

Let me know if that helps or not.

superjax:
Both of those things have been done.  I actually pushed the updated mavlink into the branch

superjax:
Okay, it looks like I was going too fast.  It can handle 100 Hz okay without messing up anything else, any faster than that and the other mavlink messages slow down.

PR #21, opened 2016-06-16
Rc

superjax:


dpkoch:
Wow, you've done a lot! It looks awesome.

I'm still working through the muxing logic to make sure I understand what's going on; I'll try to finish that as soon as possible.

In the meantime though, I did notice one issue with a few of the parameters you added. The max parameter name length is 16 characters (not counting the null termination character; you get that for free), which is limited by the mavlink parameter interface. So... we need to shorten some of the longer names. The enum names can stay long if you'd like, but the actual string name needs to 16 bytes or less.

superjax:
This commit was just tested with Gazebo.  I can confirm that the autopilot code was trying to pilot the airplane.   I am going to try to go fly this with Dallin tomorrow.

src/mode.c
```diff
@@ -0,0 +1,106 @@
...
+
+  // check for arming switch
+  if (_params.values[PARAM_ARM_STICKS])
+  {
+    if (_armed_state == DISARMED)
+    {
+      // if left stick is down and to the left
+      if (pwmRead(_params.values[PARAM_RC_F_CHANNEL]) < _params.values[PARAM_RC_F_BOTTOM] +
+          _params.values[PARAM_ARM_THRESHOLD]
+          && pwmRead(_params.values[PARAM_RC_Z_CHANNEL]) > (_params.values[PARAM_RC_Z_CENTER]+_params.values[PARAM_RC_Z_RANGE]/2)
```
dpkoch: Just double checking: is the sign on this logic correct? It seems like this is looking for the stick to the right instead of to the left to me; am I missing something?

src/mode.c
```diff
@@ -0,0 +1,106 @@
...
+        arm();
+        time_sticks_have_been_in_arming_position = 0;
+      }
+    }
+    else // _armed_state is ARMED
+    {
+      // if left stick is down and to the right
+      if (pwmRead(_params.values[PARAM_RC_F_CHANNEL]) < _params.values[PARAM_RC_F_BOTTOM] +
+          _params.values[PARAM_ARM_THRESHOLD]
+          && pwmRead(_params.values[PARAM_RC_Z_CHANNEL]) < (_params.values[PARAM_RC_Z_CENTER]-_params.values[PARAM_RC_Z_RANGE]/2)
```
dpkoch: Same thing, just wondering if the sign is correct

src/mode.c
superjax: You're right, but I was using my crappy transmitter to test this.  The default for transmitter on the rudder channel is that high is to the left, and low to the right.  On nicer transmitters, this can be switched.  I'm using a Spektrum transmitter now so we should probably have it the way we expect.  We should also get rid of the mavlink debug.

src/mode.c
superjax: Also, on the fixed_wing_test branch (where most of my work has been focused the last two days) I added the ability to switch the direction on the servo output.  Does it make sense to add the ability to switch input directions, or does that just make it confusing?

src/mixer.c
```diff
@@ -32,11 +36,11 @@ static mixer_t quadcopter_x_mixing =
 
 static mixer_t fixedwing_mixing =
 {
-  {M, S, S, S, 0, 0, 0, 0},
+  {S, S, M, S, 0, 0, 0, 0},
```
dpkoch: Just curious what motivated this ordering? My initial thought would be to do aileron, elevator, rudder, throttle. Are these the RC channels or something? If it's a decision between matching RC inputs or making the offboard command setpoints intuitive, I would probably go with the latter, even if it means adding some remapping logic in the code here. Thoughts?

dpkoch:
This is very impressive! To be honest, it's more code than I can really go through and understand the details of at the moment, so I'm just going to trust you on most of it. At some point we'll probably want to do a high level code review and make sure that the architecture is all set up how we want and is as clean as we'd like it. Chances are we'll need to change a few things, but I think this is a great start.

Here are a few comments on things I did notice:
- It would probably be good to abstract out a lot of the RC stuff. So instead of checking PWM values in all the if statements, we would just get a normalized value from an RC module that takes care of all the calibration stuff under the hood. We probably also need to revamp the RC calibration parameters; we might need something like min, max, trim, and direction for each channel.
- It might be a good idea to move the arming detection stuff in mode.c to a state machine based approach, just so it's easier to document what's going on there
- I suspect the data flow through the muxing and controller pieces could be streamlined a bit, but I'd have to think about how for a while. I think this looks great for now.

With that said, I'm good with merging as is, we'll just want to make sure we create issues for whichever of these points you agree with so they don't get lost. Thoughts?

superjax:
I agree with all of the above. 

1.   The reason I did the manual PWM check is purely motivated by speed.  I wanted to try to get a flight test as soon as possible, I expect that eventually we should have a "check switch" type of method which does the PWM checking, direction, etc.. which is sensitive to user-assignable parameters.  What I have is a first hack.  A working first hack, but a first hack nonetheless

2. I agree that a state machine would be easier to document, however I wasn't entirely sure how to actually implement it.  I'm totally open to a revamp if you're interested.  (I also think that we should have the yaw direction assignable via a parameter, so we could abstract the RC reading in a similar manner to the PWM stuff we talked about in 1.

3 - So, if you'll compare the code with my flow chart, you'll see that it doesn't agree.  The flow chart doesn't work if the RC channels are `active` from stick deviations.  (Basically, my original flow design wasn't right, so I made some modifications to make it right.)  It might be more efficient to re-design the flow, then code it again.

> Make it work, Make it right, Make it fast.
Kent Beck

PR #30, opened 2016-06-23
Estimator

superjax:
Quaternion-based estimator.  Currently using floats.  The full estimator takes about 190 us currently.

I did some work also on fixing the stream rates, so that they match what we specify in the params, and changing the baud rate via a param.


lib/turbotrig/turbovec.c
```diff
@@ -0,0 +1,238 @@
...
+  // xyz = s1*v2 + s2*v1 - v1 x v2)
+  intvec_t xyz = int_vector_sub(int_vector_add(int_scalar_multiply(s1, v2), int_scalar_multiply(s2,v1)), int_cross(v1,v2));
+  intquat_t q = {w, xyz.x, xyz.y, xyz.z};
+  return int_quaternion_normalize(q);
+}
+
+
+intquat_t int_quaternion_inverse(intquat_t q)
+{
+  intquat_t out = {-1*q.x, q.y, q.x, q.z};
```
dpkoch: Is this right? I would expect `out = {q.w, -q.x, -q.y, -q.z }`?

superjax: Either works.  One is reversing the angle, the other reverses the vector. I haven't actually tried both ways, but I expect that it would result in the same thing

dpkoch: Okay that's probably fine (although it seems a bit weird to me; reversing the angle would still result in {w,-x,-y,-z} because the w is cos(theta/2) which is even, while the x,y,z are multiplied by sin(theta/2) which is odd).

My bigger concern though, is the indices. In your case I would expect {-w,x,y,z}, but you've got {-x,y,x,z}  which seems a bit wonky?

superjax: Oh, you're right.  Hahah.  I didn't notice that, I haven't actually used the int_quaternion yet.  I have just used the float-based stuff.  I'll go ahead and make that change.

I'll switch it to `-x, -y, -z` as well,

dpkoch:
This is going to be great! I think it's pretty awesome you've dug into this and come up with a really nice peer-reviewed solution. I think it'll help give us a lot of credibility, and hopefully it'll perform really well too.

One question: are you planning on sticking with the float stuff then? If it's fast enough then I'm fine with it (it's a lot easier to read that way at least), but I'm just wondering what your thoughts are there.

I think this is great, but before we merge it in I do think there are a few things to clean up though. I'm going to start making comments and/or changes, so please don't hate me...

lib/breezystm32
```diff
@@ -1 +1 @@
-Subproject commit fb14d700999218e09a1f089847924324908766b9
+Subproject commit 3c404abcd6546fc19846bbe4a60011ebed7c3784
```
dpkoch: So it looks like you've got an outdated BreezySTM32 reference that got committed. Not hard to fix, but I think that's happened a couple of times so maybe just try to keep an eye out for that. I'll go ahead and fix this one.

lib/mavlink/v1.0
```diff
@@ -1 +1 @@
-Subproject commit 6251d3f235d653e88eec8f599787adfbd361f358
+Subproject commit fc2580177a4e2422d4f7f38b5c59a100064b1fe8
```
dpkoch: @superjax So there's a script that's part of the mavlink repo ([update_c_library.sh](https://github.com/BYU-MAGICC/mavlink/blob/master/scripts/update_c_library.sh)) for autogenerating the headers and pushing them to the `mavlink_c_library` repo. I think for consistency, and to avoid problems like missing files, it'd be better to use that rather than adding in the new files manually. I've done that and updated this submodule reference to point to the new commit.

lib/turbotrig/turbovec.c
```diff
@@ -98,7 +98,7 @@ intquat_t int_quaternion_multiply(intquat_t q1, intquat_t q2)
 
 intquat_t int_quaternion_inverse(intquat_t q)
 {
-  intquat_t out = {-1*q.x, q.y, q.x, q.z};
+  intquat_t out = {q.w,-1*q.y, -1*q.x, -1*q.z};
```
dpkoch: So close! Still have x and y mixed up lol...

src/main.c
```diff
@@ -52,14 +73,16 @@ void setup(void)
   init_mixing();
 
   // Initialize Estimator
-  init_estimator();
+  init_estimator(false, true, true);
   init_mode();
   _armed_state = ARMED;
```
dpkoch: Is this a debug thing we should get rid of?

src/main.c
```diff
@@ -76,17 +99,25 @@ void loop(void)
...
+    counter++;
 
-    // If I have new IMU data, then perform control
-    run_estimator(now);
-    run_controller(now); // 6us
-    mix_output();
+    //    // If I have new IMU data, then perform control
+    run_estimator(now); // 234 us (acc and gyro, float-based quad integration, euler propagation)
+    //    run_controller(now); // 6us
+    //    mix_output();
```
dpkoch: Do we want these lines uncommented for merging back into master?

src/mavlink_stream.c
```diff
@@ -24,6 +25,11 @@ static void mavlink_send_heartbeat(void)
                              MAV_STATE_STANDBY);
 }
 
+static uint32_t time_mavlink_send_message(void* f)
```
dpkoch: This is new to me; are we actually using this? If not perhaps we should get rid of it

src/mavlink_stream.c
```diff
@@ -96,16 +102,16 @@ void mavlink_stream(uint32_t time_us)
   {
     if (mavlink_streams[i].period_us && time_us - mavlink_streams[i].last_time_us >= mavlink_streams[i].period_us)
     {
-      mavlink_streams[i].last_time_us = time_us;
+      // if we took too long, set the last_time_us to be where it should have been
+      mavlink_streams[i].last_time_us = time_us - (time_us - mavlink_streams[i].last_time_us-mavlink_streams[i].period_us);
```
dpkoch: I see; I think that makes sense. However, should we just simplify this to
```
mavlink_streams[i].last_time_us += mavlink_streams[i].period_us;
```
?

Also, is it possible that doing this would cause a too-fast stream request to saturate the bandwidth (we'll always take too long, so it'll always be trying to send)? Although that may happen the old way too. I guess we can try it this way for now and I'll look into it when I finally dig into the mavlink timing stuff.

src/mavlink_stream.c
```diff
@@ -96,16 +102,16 @@ void mavlink_stream(uint32_t time_us)
...
       mavlink_streams[i].send_function();
     }
   }
 }
 
 void mavlink_stream_set_rate(mavlink_stream_id_t stream_id, uint32_t rate)
 {
-
-  mavlink_streams[stream_id].period_us = (rate == 0 ? 0 : 1e6 / rate);
+  mavlink_streams[stream_id].period_us = (rate == 0 ? 0 : (int32_t)((1000000.0f) / (float)rate));
```
dpkoch: So... maybe I'm just being stupid and missing something, but I don't think this helps us at all. Converting to floats before doing the division will mean the remainder gets put into the stuff after the decimal point, but then casting back to an int will get rid of all that anyway. Isn't it equivalent to integer division?

src/param.c
```diff
@@ -43,17 +43,19 @@ void set_param_defaults(void)
...
   }
-  init_param(PARAM_BOARD_REVISION, "BOARD_REV", 5);
+  init_param(PARAM_BOARD_REVISION, "BOARD_REV", 4);
+
+  init_param(PARAM_BAUD_RATE, "BAUD_RATE", 921600);
 
   init_param(PARAM_SYSTEM_ID, "SYS_ID", 1);
   init_param(PARAM_STREAM_HEARTBEAT_RATE, "STRM_HRTBT", 1);
-  init_param(PARAM_STREAM_IMU_RATE, "STRM_IMU", 100);
+  init_param(PARAM_STREAM_IMU_RATE, "STRM_IMU", 500);
```
dpkoch: Do we want the default that high? I'd think we should keep it at 100 since that's sufficient for most applications and 500 might clog up the line?

dpkoch:
So I'm all about unit tests, and I think this is a great start. However, ideally a unit test would be self-contained so that it can get run and automatically report a pass/fail during the Travis build. Then those tests would have to pass as well as the simple build test we're using now before we would approve a merge. This can really help with regression testing; making sure new code doesn't break existing functionality. Is there some way we can refactor this one for that to happen?

With that said, I think we're going to need to actually design a test framework at some point (a defined way for us to add and run unit tests with as much automation as is possible/reasonable). That's going to take some work. Do you think we should merge this one in for now, or wait until we figure that out?

lib/turbotrig/turbovec.c
```diff
@@ -0,0 +1,240 @@
...
+float turboInvSqrt(float x)
+{
+  long i;
+  float x2, y;
+  const float threehalfs = 1.5F;
+
+  x2 = x * 0.5F;
+  y  = x;
+  i  = * ( long * ) &y;                       // evil floating point bit level hacking
+  i  = 0x5f3759df - ( i >> 1 );               // what the f***?
```
dpkoch: Haha yeah... as funny as I think this is, perhaps we should reword it?

src/mavlink_stream.c
superjax: Sorry, I forgot to delete this.  It is a leftover from one of my attempts and correcting the mavlink send timing.

src/mavlink_stream.c
superjax: So far, I haven't had that problem.  I had it streaming at 1000 Hz, and it wasn't any problem (any faster is stupid, because the IMU only reports at 1000 Hz).  The param reporting at the beginning slowed down, but I was still sending commands, receiving baro and mag, everything.  I think the mavlink-based system gives us _way_more efficient use of the bandwidth as compared with the old MSP system.

src/mavlink_stream.c
superjax: I think it is the same.  Again, I believe this float stuff is leftover from my other attempts at being smart about streaming.

src/param.c
superjax: I haven't had any problems at 500 Hz. Like I said above, 1000 Hz was doing great, simultaneously streaming commands at full rate, baro, differential pressure, and a `named_value_int` at something like 100 Hz.  I think we have a lot more bandwidth on this new system.  I think it would be a good idea to try to push the limits and see where stuff starts to break, but it's certainly not at 500 Hz.

superjax:
I agree with you on the unit tests.  This is a total first hack at a unit test (a proper test would include coupling between axes, maybe an rk4 'oracle' to test against, you know, the whole thing.  This is a helpful test, but it isn't automated.  I believe that it should be pretty easy to automate given some time.

I'm not actually sure how Travis builds work.  I could set up a way that `main.c` could return a `true` or `false`.  Would you be able to use in Travis tests?  I have a python "oracle" that I used to write the prototypes.  Is there a way we could use it to test this code?  It has an rk4 implementation that provides truth in coupled axes movements, and it can give you an RMS error for the estimator.  I don't know if we can use it at all, but it might be useful.

dpkoch:
Hey just FYI, it looks like you pulled in outdated submodule references again in your last commits. It's not the end of the world, but it's happened a few times now so I'll try to give you a quick tutorial on it:

Basically what happens is a change will get pushed/merged that changes one of the submodule commits. When you pull those changes down to your computer, git does _not_ automatically update your local submodule references to match. Therefore your outdated references show up as "modified" in your git status (the submodule path will be listed) and a git diff will show you which commit hash you have locally vs what's in the repo. To fix it you have to run `git submodule update` to update your local references to the new ones.

So what happens is that if you pull in a change that modifies the submodule references, make some changes, and then add everything with a `git add .` or some similar command, your outdated submodule references get added as something you want to push to the repo. So just be aware of that, and try to keep an eye out when you run your git status/diff to stop that from happening in the future.

src/param.c
dpkoch: Sounds good; let's keep it for now at least.

dpkoch:
Sounds good about the floating point stuff. I just wasn't sure what your thoughts were since you are Mr. Integer :)

But that's great! That seems plenty fast, and I think it's definitely a lot more readable/maintainable in floating point. At some point, if we decide to stick with it permanently, we may want to think about making things a bit more uniform across the board (e.g. do accel and gyro values as floats in normal units) if the performance hit is small but the readability is a lot better.

superjax:
Let's put it on a different branch until we have a good way of doing unit tests.  I'll talk to Robert Pottorff about doing unit tests, I know that he's Mr. auto tester so I'm sure he'll have some good ideas.

Sorry again about the gitmodule stuff.  Thanks for the tutorial, that should help me not make a mess of it any more.

superjax:
Alright, so I think it should be good to go, but something is still weird with the submodules.  I think after this pull request, I'm going to blast away my local repo and pull it again, because I think I've messed up my submodule references.  (I did, however get the temperature working on the MPU6050 and Simon pulled it in)

PR #34, opened 2016-06-30
Time sync

### LINKED ISSUE #11: IMU Mavlink
Hey @dpkoch,

I've been thinking a lot about trying to get good IMU timestamps.  I talked with Dr. McLain, and he pointed out that we can often make do with just knowing the difference between measurements, rather than the absolute timestamp.  I really liked that idea, and I'm wondering if you could add a field to the IMU message (could be just int16), for a time stamp.  It would be defined as the number of microseconds since the last IMU measurement was taken that this measurement was taken.

Simon pulled my changes to Breezy so we have an interrupt routine to play with to get the IMU timing data, so it should be pretty straight-forward.

dpkoch:
Implemented time synchronization modeled after mavros and the PX4.

Time synchronization currently works for message types that include a time stamp. It would also be possible to estimate the delay and subtract that off of the time other, non time-stamped messages are received to make the resulting ROS time stamp a little more accurate, but I have not done that yet.

Closes #11.

### DIFF
--- a/src/mavlink_receive.c
+++ b/src/mavlink_receive.c
@@ -64,6 +64,19 @@ static void mavlink_handle_msg_command_int(const mavlink_message_t *const msg)
   }
 }
 
+static void mavlink_handle_msg_timesync(const mavlink_message_t *const msg)
+{
+  uint32_t now_us = micros();
+
+  mavlink_timesync_t tsync;
+  mavlink_msg_timesync_decode(msg, &tsync);
+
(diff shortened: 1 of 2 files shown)


superjax:
This is awesome!  Thanks!  I'm going to wait until the comments in BYU-MAGICC/mavlink#3 is closed before merging

dpkoch:
Alright it's all in, except mavlink just streams a zero for temperature right now

superjax:
Okay, I'll get that going, you'll be seeing a pull request later today

Issue #32, opened 2016-06-27
IMU Accelerometer Bias Compensation


superjax:
We need a good way to make the IMU calibration better.

Ideas:
Temperature Compensation? 
[http://www.i2cdevlib.com/forums/topic/53-mpu-6050-temperature-compensation/](url)

Using the DMP?
[motion_driver_6.12.zip](https://github.com/BYU-MAGICC/ROSflight2/files/335917/motion_driver_6.12.zip)

superjax:
I've got the temperature reading in MPU6050.  I'll pull it into Simon's library soon.

It appears that the procedure for doing temperature compensation is taking a bunch of measurements of the bias at different temperatures, then finding a least-squares approximation of the bias vs. temperature for all the axes.  We already have the `correct_imu` function in `sensors.c.` I think that it wouldn't be too bad to throw in a temperature compensation term on each axis.  It seems that most people use a linear compensation, while some people have gone for quadratic.  I'm not sure which we should do, but I'll get started on implementing that.

superjax:
@dpkoch.  Do you have an automated calibration already set up?  If you do, I expect that it would be faster for me to modify your calibration code than write my own.

dpkoch:
Yeah we definitely need to do something about the calibration, and I think temperature compensation is a big part of it. I just started with a room temperature Naze (unplugged), then plugged it in and recorded IMU data for an hour while it sat on the counter. Check out the plots:
![accel](https://cloud.githubusercontent.com/assets/2702490/16427304/552ff6d2-3d32-11e6-9d44-8dbc9cd78a97.png)
![gyro](https://cloud.githubusercontent.com/assets/2702490/16427303/552f0d30-3d32-11e6-98dd-093ac87ad89d.png)

So the biases definitely drift over time, which I assume is due to change in temperature from warming up (I didn't have a way to stream the IMU temperature at the moment, but it'd be interesting to try again with that measurement). The accel z bias is particularly bad, and takes almost 20 minutes to stabilize.

The accel calibration code that I have is on my Github ([imu_calib](https://github.com/dpkoch/imu_calib)). It uses a least squares method to compute the biases, scale correction, and cross-axis misalignment terms. The least squares stuff as it stands is going to be too much for the Naze. The other issue we've been having with it is that it seems like it's pretty sensitive to having very accurate alignments (6 directions, like the pixhawk) during calibration. The lack of temperature compensation is also not helping.

[...]

Issue #35, opened 2016-06-30
New IMU data check is faulty


dpkoch:
@superjax Okay I could be wrong here, but this looks like it could be a problem.

On [line 37](https://github.com/BYU-MAGICC/ROSflight2/blob/master/src/sensors.c#L37) of sensors.c, we check the `mpuDataReady` flag to see if we have new IMU data available. However, as I was looking through drv_mpu6050.c, I realized that `mpuDataReady` never gets set to false, so after the first time that check will always return true. Is that correct, or did I miss something?

Two possible fixes:
- in sensors.c, set `mpuDataReady = false` like you do in your accelgyro example. Simple, but it seems a little weird to be modifying a flag inside a library from outside the library. Might be okay though.
- store a copy of mpuMeasurementTime and do an equality comparison. When a different value from your local copy is detected, you have a new measurement.

Thoughts? Also I wonder what this might have been doing to our timing.

superjax:
That's right. 

I just actually pulled a change to breezy where we register our own ISR - simondlevy/BreezySTM32#9.  That way, we have complete control of how the ISR works.  I got rid of the `mpuDataReady` flag entirely, and left it up to ROSflight to decide how to handle it.  It will be a lot cleaner, and it works better for camera triggering.

superjax:
testing

dpkoch:
Sweet! That'll be really nice.

dpkoch:
Closed by #36

PR #36, opened 2016-06-30
Temperature

superjax:
Added Temperature support on IMU message
Added Custom ISR


Makefile
```diff
@@ -40,7 +40,7 @@ OPTIONS		?=
 DEBUG ?=
 
 # Serial port/device for flashing
-SERIAL_DEVICE	?= /dev/ttyUSB0
+SERIAL_DEVICE	?= /dev/ttyUSB1
```
superjax: This shouldn't have been added

src/sensors.c
```diff
@@ -11,34 +12,50 @@ int16_t _accel_data[3];
...
     mpu6050_read_accel(_accel_data);
     mpu6050_read_gyro(_gyro_data);
-    correct_imu();
+    mpu6050_read_temperature(&_imu_temperature);
+
+    // correct according to known biases and temperature compensation
+    _accel_data[0] -= (_params.values[PARAM_ACC_X_TEMP_COMP]*_imu_temperature)/1000 + _params.values[PARAM_ACC_X_BIAS];
+    _accel_data[1] -= (_params.values[PARAM_ACC_Y_TEMP_COMP]*_imu_temperature)/1000 + _params.values[PARAM_ACC_Y_BIAS];
+    _accel_data[2] -= (_params.values[PARAM_ACC_Z_TEMP_COMP]*_imu_temperature)/1000 + _params.values[PARAM_ACC_X_BIAS];
+    _gyro_data[0] += _params.values[PARAM_GYRO_X_BIAS];
```
dpkoch: Should be a minus sign? (subtract the bias off)

superjax: Yep!  You're right.

src/main.c
```diff
@@ -20,12 +20,17 @@
...
+serialPort_t *Serial1;
+
 int main(void)
 {
   // Configure clock, this figures out HSE for hardware autodetect
   SetSysClock(0);
   systemInit();
 
+  // Initialize Serial ports
+  Serial1 = uartOpen(USART1, NULL, 921600, MODE_RXTX);
```
dpkoch: So it looks this removes support for setting the baud rate via a param. Just checking, is that what you intended?

superjax: Oh.  Shoot.  no, that's not what I meant to do.

dpkoch:
Cool, I think we're close here.

Here's how I envision the data flow happening:

1. When we do a bias calibration, the result gets stored in the params like you're doing now
2. When we run `update_imu`, we apply the stored accel bias and temperature compensation (which you're doing already) and the stored gyro biases (which isn't happening right now) to the values stored in `_accel_data` and `_gyro_data`
3. The estimator uses those compensated values as its inputs, but then also internally keeps track of an additional bias term to further correct it. Those bias terms get stored in a global variable, but not in the parameter value; that stays where it was from the calibration.
4. For the mavlink streaming, by default we stream the values corrected by the param stuff (already stored in the global `_accel_data`, `_gyro_data`). If we choose to apply the estimator corrections, then we additionally subtract those biases off.

Does that seem reasonable/in line with what you're thinking?

superjax:
I think that'll work.  Sure, let's do that.  We'll have an adaptive gyro bias calculation, and the static bias.  Cool!

superjax:
What do you think of it now?

dpkoch:
Sweet! Looks great. I'm just going to test it with fcu_io and run the style script real quick, then I think we're good to go!

dpkoch:
Going along with BYU-MAGICC/fcu_io2#24, let's get this guy wrapped up! Again, we still just need to decide if we want to go with floating point, in which case we'd merge in the temperature_float branch.

dpkoch:
This pull request will close #32, #35, and #37

superjax:
Alright, I'm going for it!

PR #42, opened 2016-07-07
[WIP] Make onboard bias calibration play nice with temperature calibration

dpkoch:
This fixes an issue where the bias offset from the onboard calibration is not compatible with the offset from the temperature calibration.

The reason is that the temperature calibration computes what the offset would be at 0degC, while the onboard calibration computes it at whatever the current temperature is. There are two possible ways to fix this:

1. Change the temperature compensation to use a bias offset computed at an arbitrary value (so instead of `bias = bias_0 + m*T`, we would have `bias = bias_c +  m*(T - T_c)`. This is probably the more intuitive solution, but requires storing another parameter and an additional subtraction at each time step. Also requires a slight tweak to the least squares stuff, but it's not bad at all.
2. Modify the onboard bias calibration to estimate the bias at temperature 0 based on the current slope (so instead of `bias_0 = bias_c`, we have `bias_0 = bias_c - m*T_c`). Less intuitive, but eliminates the need to store `T_c` and the extra subtraction is done only once.

I've implemented option 2 here, and it seems to work pretty well. Running the temperature calibration then the onboard calibration immediately after results in the same offsets, where before they would change. I'm also open to doing 1 instead if you think the clarity is worth the cost.

With that said, the z offset still isn't quite right. I come in at about 9.6m/s^2 instead of 9.8. Would you mind testing yours for comparison?

### DIFF
--- a/src/sensors.c
+++ b/src/sensors.c
@@ -67,19 +67,22 @@ static bool update_imu(void)
     {
       static int16_t acc_count = 0;
       static int32_t acc_sum[3] = {0, 0, 0};
+      static int32_t acc_temp_sum = 0;
       acc_sum[0] += _accel_data[0];
       acc_sum[1] += _accel_data[1];
       acc_sum[2] += ((_accel_data[2]*_accel_scale)-9806650)/_accel_scale;
+      acc_temp_sum += _imu_temperature;
       acc_count++;
       if (acc_count > 100)
       {
-        _params.values[PARAM_ACC_X_BIAS] = acc_sum[0]/acc_count;
-        _params.values[PARAM_ACC_Y_BIAS] = acc_sum[1]/acc_count;
-        _params.values[PARAM_ACC_Z_BIAS] = acc_sum[2]/acc_count;
+        _params.values[PARAM_ACC_X_BIAS] = (acc_sum[0] - _params.values[PARAM_ACC_X_TEMP_COMP]*acc_temp_sum/1000)/acc_count;
+        _params.values[PARAM_ACC_Y_BIAS] = (acc_sum[1] - _params.values[PARAM_ACC_Y_TEMP_COMP]*acc_temp_sum/1000)/acc_count;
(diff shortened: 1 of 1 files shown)


superjax:
So I am also getting it right at 9.6.  However, if I flip it over, It goes to 9.8.

Just playing around, the results of my calibration gave me this:
```
[ INFO] [1467987039.384289327]: IMU temperature calibration complete:
 xm = 0.004180, ym = -0.050857, zm = -0.098083 xb = 387.603525 yb = -293.292131, zb = -41.777126
```
So if i manually entered my z-bias to be -42, It went to 9.6.  If I typed it in as -84, it went to the correct value of 9.8.

Still looking into it, but that's what I have to report so far.

superjax:
I fixed it - It was a problem with fcu_io

In `imu.cpp` line 62, it used to be
```c
measurement << msg.xacc, msg.yacc, msg.zacc -4096; //! \todo need a better way to know the z-axis offset
```

I changed it to be 
```c
measurement << msg.xacc, msg.yacc, ((msg.zacc*ACCEL_SCALE) - 9.80665)/ACCEL_SCALE;
```
I accidentally just pushed to master though.  Do you have any problem with my fix?

dpkoch:
Okay that's good. So that should fix the temperature bias calibration then (I haven't tested, but you can confirm that the temperature calibration makes it go to 9.8?).

However, the onboard calibration is still broken; I still get 9.6 when I run that one. I was thinking line 73 in sensors.c might be suspect due to integer rounding or something:
```
acc_sum[2] += ((_accel_data[2]*_accel_scale)-9806650)/_accel_scale;
```
However, if I change it to be
```
acc_sum[2] += _accel_data[2] - acc1G;
```
where `acc1G` is the value returned by `mpu6050_init` (or hard-coded to 4096; I tried both), I still get 9.6. Any thoughts?

dpkoch:
Also, are you still getting 500Hz IMU messages with the current code? Mine has slowed down to ~425Hz for some reason.

superjax:
no, I'm still getting ~499

superjax:
I tried streaming it at 1000, and it had slowed down to 970, however, I turned off servo and rc streaming, and it got back up to 997.  I just realized that we don't have a way to adjust the baro, airspeed, mag and sonar rates, so I've started working on that.

Issue #29, opened 2016-06-20
IMU scaling depends on board revision


dpkoch:
@superjax So I flashed the latest firmware on both a rev5 and a rev2 Flip32 board. When streaming IMU over mavlink, the accel scale factor in fcu_io2 rev5 board is correct (I get 9.8m/s^2 at rest). The scale factor is off by 2 (I get 4.9m/s^2 at rest) for the rev2 board.

The solution to this problem may depend on fcu_io2, but it's also possible that this could be affecting onboard estimation. On the fcu_io2 side we could imlement some way to retrieve the scale factor from the naze. Another possibly is to send scaled values over mavlink instead, although I'm not sure what the answer is there  based on our previous discussion.

Either way, we'll need to ensure that the accel is being scaled correctly on board when necessary (e.g. for altitude estimation). We may already have what we need from the mpu6050_init, but we'll need to check.

superjax:
This should be linked to BYU-MAGICC/fcu_io2#24

dpkoch:
I have confirmed that `_accel_scale` is at least close to correct for both the rev2 (4807) and rev5 (2394) boards. The rev5 value matches what we've been using in fcu_io. I would expect them to be off by a factor of 2, but they're not quite, so that's a little bit weird.

superjax:
This is definitely a problem with `mpu6050_init' in Breezy.  I believe that it has something to do with the 'half' business going on in the initialization.

superjax:
The true `_accel_scale` should be 9.807/16384 or 0.00059857177.  This should be true for both boards.  They both have a 8G scale on a 16 bit ADC.

superjax:
I may have fixed a bug in `mpu6050_init`.  Here is the relevant parts of the datasheet
![image](https://cloud.githubusercontent.com/assets/7344014/16609373/f2dfd9b8-4310-11e6-9da4-fd3778e49cfc.png)
![image](https://cloud.githubusercontent.com/assets/7344014/16609382/0587fb9a-4311-11e6-993f-4e2dd595e87b.png)
and here is the original code
```c
enum accel_fsr_e {
    INV_FSR_2G = 0,
    INV_FSR_4G,
    INV_FSR_8G,
    INV_FSR_16G,
    NUM_ACCEL_FSR
}; 
...
mpuWriteRegisterI2C(MPU_RA_ACCEL_CONFIG, INV_FSR_8G << 3);
```
[...]

PR #48, opened 2016-07-11
Temperature float

### LINKED ISSUE #38: Move towards floating point *gasp* data representations?
Hey @superjax 

So I'm just wondering if we want to take another step down the dark path you started us on by implementing your estimator with floating point arithmetic.

My question is this: should we just go for it and do all of our sensors (especially IMU) in floating point too? Here are some reasons why I think it may be a good ideal:
 - Doing the estimator in floating point didn't seem to be too much of a performance hit, so this might not either
 - Whatever performance we do lose, we could regain by going to the F3. Maybe loops run slower on the lesser boards, but if anybody wants serious performance they can upgrade
 - It would make the code __much__ more readable
 - I think it would simplify a lot of the issues we're seeing with getting this calibration stuff working. Right now we're trying to stream and calibrate in terms of raw values, but something seems to be getting lost in the translation. If we were just streaming everything over in SI units, figuring out some calibration parameters, then applying those on the Naze as SI units, we'd be much less likely to screw something up.

Obviously, the downside would be a speed hit.

At this point I'm just asking the question. What are your thoughts?

dpkoch:
Switch IMU data to floating point, both internally and over Mavlink. Closes #38, helps with #29.

Before merging, need to test the following:
- [x] parameter get/set interface for both integer and floating point values
- [x] accelerometer calibration, both in fcu_io and in ROSflight

I have to run but can do those tests in about an hour.


dpkoch:
Tested and should be good to go.

Removed default temperature compensation because they seemed to be off. This might be a unit change thing, but it also seemed like the values were not consistent between units. May need to do more evaluation to come up with a good default value.

superjax:
Awesome!

PR #49, opened 2016-07-12
Added printf like support for mavlink STATUSTEXT messages

### LINKED ISSUE #19: MAVLink status text
We need to be able to send messages (text) from the fcu to be displayed by fcu_io. MAVLink has the [`STATUSTEXT`](https://pixhawk.ethz.ch/mavlink/#STATUSTEXT) message for this purpose, so we just need to implement that.

We may need to implement some slightly fancier buffering onboard so these don't clog up the line. Will look into it.

dpkoch:
Closes #19.

One warning though: these statustext messages are big and don't get buffered onboard, so use them sparingly!

### DIFF
--- a/include/mavlink_log.h
+++ b/include/mavlink_log.h
@@ -1,0 +1,16 @@
+#pragma once
+
+#include <breezystm32/breezystm32.h>
+
+#include "mavlink.h"
+
+#define mavlink_log(severity, format, ...) do { \
+  char text[MAVLINK_MSG_STATUSTEXT_FIELD_TEXT_LEN]; \
+  sprintf(text, format, ##__VA_ARGS__); \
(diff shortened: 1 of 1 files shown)


dpkoch:
I know it's kind of ugly with all of the macro stuff, but it's the only way to do it without rewriting the `sprintf` function. You have to expand out the variable argument list via macro as you pass it into sprintf, otherwise you'd need a version of sprintf that accepts a `va_list` argument which we currently don't have. So it's ugly, but it's still probably our cleanest option.

Also, the `printf`/`sprintf` function packaged with BreezySTM32 doesn't support floating point values, so that's kind of a bummer. But hey, it's something at least.

superjax:
Dude, this is awesome.  I really really like how you tied it in with the ROS default print statements, it makes it really clean and natural.

I just tested it out, and it's super.  MERGED

Issue #23, opened 2016-06-17
HIL Simulation


superjax:
I decided to test the RC muxing and offboard control stuff with Gazebo.  I figured that it would be pretty easy to just grab the command messages coming off the autopilot and push them to the naze.  It was _incredibly_ easy.

I'm wondering if we should make this a supported feature.  I suppose I'll have to play around with bandwidth, which is the main concern, but I think that it wouldn't be too terrible to support a HIL simulator for ROSflight and Gazebo, certainly for fixed-wings it shouldn't be too much of a problem, but I'm thinking that it actually wouldn't be too crazy to get it going for the multirotors too. 

Stuff we would have to do.

- [x] Forces and Moments would need to subscribe to the `servo_output_raw` topic, rather than `command` like it currently does

- [x] IMU would need to be streamed _into_ the naze, rather than out of it.

That doesn't sound too bad.  I am certain that there is enough bandwidth to handle the fixed wing information.  I'm not so sure about the multirotor (in fact, I'm not even sure that we can get Gazebo running at a full 1000 Hz to simulate the multirotor)  Something to think about.

superjax:
So, I ultimately decided to not do this.  The latency is too great over mavlink to perform accurate simulation.  Software in the loop is a lot more reliable and realistic.

PR #50, opened 2016-07-13
[WIP] HIL

superjax:
I got the hardware in the loop simulation working (in preparation for building the controller)  Right now, the controller is wacked, but the HIL seems to be working.

We can't know for sure that the HIL is working until the controllers are operating properly, but I wanted @dpkoch to look at it before I got too far.

Addresses #23


src/mavlink_receive.c
```diff
@@ -100,6 +100,27 @@ static void mavlink_handle_msg_timesync(const mavlink_message_t *const msg)
   }
 }
 
+static void mavlink_handle_msg_small_imu_hil(const mavlink_message_t * const msg)
```
dpkoch: Is a `SMALL_IMU_HIL` message any different from a `SMALL_IMU` message? If not, do we want to just use the existing message type, or did you have future changes in mind?

superjax: The only difference is that the `SMALL_IMU_HIL` doesn't have temperature

superjax: I also didn't wan't it to be confusing, so I figured it made it more clear to have two types

src/sensors.c
```diff
@@ -46,7 +46,19 @@ void imu_ISR(void)
 
 static bool update_imu(void)
 {
-  if (_imu_ready)
+  if (_params.values[PARAM_HIL_ON])
+  {
+    // don't update normally, use HIL values
+    if(_imu_ready)
```
dpkoch: So it looks like this still uses the timing from the MPU6050 interrupt rather than the timing from the incoming HIL IMU messages. Is that going to be a problem? Do we want to have the incoming IMU message event set a flag that `update_imu` checks instead of `_imu_ready` if the `PARAM_HIL_ON` param is set?

superjax: I was actually intending to use the same flag, but I forgot to de-register the imu interrupt.

src/controller.c
```diff
@@ -22,7 +25,7 @@ int32_t sat(int32_t value, int32_t max)
 
 void run_controller(uint32_t now)
 {
-  control_t outgoing_command = rate_controller(attitude_controller(_combined_control, now), now);
+  control_t outgoing_command = rate_controller(_combined_control, now);
```
superjax: Until I have the rate controller working, I'm just ignoring the attitude controller

src/mavlink_receive.c
dpkoch: Cool, I'm good with that.

superjax:
So, turns out there is too much latency to actually fly a multirotor in HIL, that or the timing isn't working out well.  Here is the response of the naze's estimator vs. truth for a multirotor flight.  The naze was working in HIL.

image

It's okay.  Not great, but okay.  There seems to be a weird timing issue.  My terminal was full of these 
```
[ INFO] [1468592372.320812514, 134.002000000]: Detected time offset of -26.623121000 s
[ INFO] [1468592373.396364383, 135.001000000]: Detected time offset of -26.696940000 s
[ INFO] [1468592374.453129385, 136.001000000]: Detected time offset of -26.752051000 s
[ INFO] [1468592375.488622909, 137.002000000]: Detected time offset of -26.785315000 s
```

Which makes me think that there is too much latency to do accurate simulation.

Any ideas?

superjax:
Turns out, it's not entirely a latency issue, but it is bad enough that I don't think it's worth it.

PR #51, opened 2016-07-15
Software In-The-Loop

superjax:
This sets up ROSflight to be used within a Gazebo Library.  I also did some fixing of the controller, moved it to float etc...


src/sensors.c
```diff
@@ -155,7 +160,9 @@ void init_sensors(void)
   mpu6050_register_interrupt_cb(&imu_ISR);
 
   uint16_t acc1G;
-  mpu6050_init(true, &acc1G, &gyro_scale, _params.values[PARAM_BOARD_REVISION]);
+  float gyro_scale_to_urad;
+  mpu6050_init(true, &acc1G, &gyro_scale_to_urad, _params.values[PARAM_BOARD_REVISION]);
+  gyro_scale = 1000000.0*gyro_scale_to_urad;
```
dpkoch: Should this not be `gyro_scale = gyro_scale_to_urad / 1e6;`? (you have to divide by 1e6 to get from urad to rad, not multiply, right?)

dpkoch:
That's pretty sweet; so the ROSflight code is what Gazebo uses as the attitude controller now? So essentially Gazebo becomes a Naze emulator with SIL, am I understanding that correctly? If so then that's pretty awesome!

So we've got a few more things to do for changing over to the float stuff I think:
- Change `_current_state` to floats with SI units. Right now we're computing attitude in floats/SI units, converting to int32, then converting back to floats/SI units in both the controller and mavlink_stream. Makes more sense to just store as a float
- Change the offboard control message to floats/SI units so everything matches
- Probably change the controller gains as float params so we don't have to keep dividing them by 1000. Would also make it a little bit easier/intuitive to tune since they'd be in normal units

However, we can merge this in before changing all of that as long as it all gets done. I've been thinking for a while that we need to parallelize our workflow more. I'm guilty of this too, but we keep creating these huge branches where we start working on one thing, but start fixing a bunch of other things on the side too so that everything is so tied together in the one branch that we can't do anything else until it's merged in. Also in cases like this, it means that if we want to abandon one direction (HIL in this case), we don't have a clean way of doing it without losing all of the other changes (arming, controller, etc.) that we want to keep.

So maybe let's merge this in fairly quickly, then try to parallelize a bit? Meaning that branches and especially commits would be more of atomic, coherent changes. If while working on one feature branch we discover a bug to fix, perhaps we should back up to master, fix the bug on a new bug fix branch and open a pull request, then pull those changes back into our feature branch so that we can keep the bug fix even if we abandon the feature. What do you think about that?

dpkoch:
Also the mavlink log throttle functions are pretty slick, but they make me a little bit nervous. Those messages are pretty huge, so I'm thinking they're intended more for one-time type messages like "calibration started/complete". Maybe the throttle versions are okay for debugging purposes, but I'd say let's not use them in any production code at least? And even for debugging use the named value messages instead if feasible?

superjax:
I agree.  They should definitely not be used in production code.  I was using them to send multiple values at once (x,y,z) at some super low rate.  It was pretty slick, but I don't intend on using them for much anymore now that I have the SIL working pretty well.  (by the way, I just flew the rate controller.  The SIL and real life are pretty similar)

src/sensors.c
superjax: The `gyro_scale_to_urad` is the mutliplier to get you _into_ urad.  Therefore, multiplying that scale factor by 1e6 results in a scale factor that gets you into rad.

dpkoch: Right, so if we start with a value in urad, then we have `x urad * (1 rad / 1e6 urad) = x rad` (you have to _divide_ by 1e6 so that urad cancels). So if `raw * gyro_scale_to_urad = x urad`, then substituting into the previous equation we get `raw * gyro_scale_to_urad * (1 rad / 1e6 urad) = x rad`, which we can write as `raw * (gyro_scale_to_urad / 1e6) = x rad`, so we would have `gyro_scale = gyro_scale_to_urad / 1e6`. Is that right, or am I just completely losing my mind?

dpkoch:
Cool. There is also a DEBUG_VECT message type (https://pixhawk.ethz.ch/mavlink/#DEBUG_VECT) we could add support for if you think that'd be useful.

src/sensors.c
superjax: The problem is that first equation.  I'll put it all in scientific notation so it makes sense

`x e-6 rad * (1 rad / 1e6) = x e-12 rad` not `x rad` .

It should be `x e-6 rad * (1e6) = x rad`

superjax:
That's a good point.  I'm still not super at trusting that a merge will work well haha.  Yeah, let's go ahead and merge this, and I can be better at being more atomic about my branching.

dpkoch:
Cool, sounds good. Maybe last thing before we merge: did you see my most recent line note up above?

superjax:
And, yes.  Gazebo uses the actual ROSflight code to perform attitude stabilization.  We used that RC benchmark dyno to get data for mikey's props, and we're modeling them a lot more accurately than before.  You might want to check out BYU-MAGICC/fcu_sim#3

src/sensors.c
dpkoch: Perhaps I'm just not understanding what `gyro_scale_to_urad` is doing. But if `raw * gyro_scale_to_urad` does actually give you a value in units of urad, then if the vehicle is rotating at a rate of 1 rad/s, `x = raw * gyro_scale_to_urad` would return a value of `x = 1e6 urad/s (= 1 rad/s)` rather than `x = 1e-6 urad (= 1e-12 rad)`. In that case we'd want to divide: `x / 1e6 = (1e6) / 1e6 = 1`. In other words, you need one million urad to equal one rad so `x` will be big; to get back to rad you have to _divide_ by one million.

?

superjax: Ah, I see what you're trying to say.  haha.

Okay. So, yeah.  You're right.  The weird thing is that it works.

dpkoch: Haha yeah, that's the weird thing about stuff like this. You find a bug that seems like it should be completely wrong, then you fix it and nothing changes. Happens to me all the time... no idea why it's like that.

So I guess the question now is will it work the "right" way?... It could be that it's actually right and we're misunderstanding what the gyro scale that `mpu6050_init` returns is doing in the first place. Or it could be that it is wrong but it can be compensated for with gains and whatnot. Difficult to say without some testing.

superjax: Yeah.  So I'm looking into it right now, I'll get back hopefully with some more insight

dpkoch: Okay, ignore that last comment. Just looking at the IMU data streaming over mavlink, it looks the units are at least close to correct as it is right now. Which means that it's probably something goofy with this line: (`drv_mpu6050.c:220`)

    // 16.4 dps/lsb scalefactor for all Invensense devices
    *gyroScale = (4.0f / 16.4f) * (M_PI / 180.0f) * 0.000001f;

I think _he's_ got his units messed up. This also goes back to your observation in https://github.com/BYU-MAGICC/ROSflight2/issues/29#issuecomment-230694862. If you get rid of the `0.000001f`, then it'd be close to actually getting you into rad, although the `4.0f` is still a mystery. But with the `0.000001f`, it puts you in units of Mrad/s? Something goofy is definitely going on there.

dpkoch: Okay I've started a little bit of research, but let's not dig into it here. Let's merge as is, and continue this investigation in #52.

superjax:
okay, sounds good

superjax:
Is there anything we need to do before merging this in?

PR #53, opened 2016-07-19
Float estimator

superjax:
Based on #51, so this pull request should be rebased on master once that request is finished.

This moves `_current_state` to be a float-based struct, represents some additional float-based turbotrig library wrappers.

I also cranked up the flash UART speed to 921600 because I'm tired of waiting forever to flash this thing.  

I tested it on my REV2 board, the estimator seems to be working great.  I found a typo in the matrix exponential method, so pitch is estimated correctly when using that method (we weren't using it previously)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,8 +177,8 @@ clean:
 	rm -rf $(TARGET_HEX) $(TARGET_ELF) $(TARGET_OBJS) $(TARGET_MAP) $(BIN_DIR) $(OBJECT_DIR)
 
 flash_$(TARGET): $(TARGET_HEX)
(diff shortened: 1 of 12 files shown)


superjax:
Thanks for cleaning up those things.  I like all your changes.  Are we good to merge?

dpkoch:
Yeah so I think the estimator is in pretty good shape now. The controller hasn't been updated to match, so it's broken right now. Do we want to clean that up as part of this pull request, or do it later?

superjax:
No, I'm changing all the command structs to floats first, then I'll get the controller working

dpkoch:
Cool. So... merge then?

superjax:
Hahah, already done.

dpkoch:
Haha just saw that. Sweet.

PR #56, opened 2016-07-19
Fixed a bunch of incorrect type castings

dpkoch:
I was messing around with compiler settings, and found a bunch of implicit type casts going on that are technically incorrect, although they probably weren't causing any problems. This fixes them though. Also using the ++ operator on an enum is technically undefined, so I got rid of those too.


dpkoch:
Just rebased after merging #55

superjax:
Awesome!

PR #57, opened 2016-07-19
Disable pedantic warnings only for mavlink_types.h

dpkoch:
Just rather than turn off the "pedantic" warnings (issued when something does not conform to the specified standard, in this case ISO C99) for the whole project, only turn them off for the offending file that we have no control over (mavlink_types.h). That way we still get warnings if we do something stupid ourselves.


superjax:
Cool!

PR #60, opened 2016-07-21
[WIP] Controller

superjax:
The controller is working on hardware.  There is a little bit of cleanup left, some that I was hoping you could help me with @dpkoch.

Both the attitude and rate loops are working from RC commands, however, I can't remember how to do integrator anti-windup.  I was wondering if you had ideas on that.   After the anti-windup is incorporated, then this branch is ready.


dpkoch:
Flight test ([video](https://youtu.be/k2K1yCY9pCA)) looked awesome!! Which mode were you flying that in?

Yep @superjax, I can get anti-windup going. One approach we could use is given in section 6.5 of the UAV book.

superjax:
That was in angle mode.  I didn't have time to really tune the rate mode, so it's still a little hairy, but it is flyable.  I wouldn't recommend using it for offboard control quite yet.  Angle mode seems to be pretty robust.

superjax:
What if we merged this as-is, and instead made integrator anti-windup another pull request?   It works as it currently stands, and it will require more testing to ensure that the anti-windup scheme is working as desired.

src/mavlink_receive.c
```diff
@@ -125,10 +125,10 @@ static void mavlink_handle_msg_offboard_control(const mavlink_message_t *const m
     _offboard_control.y.type = PASSTHROUGH;
     _offboard_control.z.type = PASSTHROUGH;
     _offboard_control.F.type = THROTTLE;
-    _offboard_control.x.value = mavlink_offboard_control.x/2.0f;
-    _offboard_control.y.value = mavlink_offboard_control.y/2.0f;
-    _offboard_control.z.value = mavlink_offboard_control.z/2.0f;
-    _offboard_control.F.value = mavlink_offboard_control.F;
+    _offboard_control.x.value = mavlink_offboard_control.x*500.0f;
```
dpkoch: So I think it'd be good at some point to move the mixer to floating point as well just so that everything can get passed through in normalized units to keep things consistent rather than carrying things around in PWM units (to me it seems like the PWM units probably shouldn't show up until as close to the actual hardware as possible). But let's make that a separate issue and not worry about it yet here.

src/mode.c
```diff
@@ -21,12 +23,14 @@ void arm(void)
...
   _armed_state = ARMED;
   LED1_ON;
+  mavlink_log_warning("ARMED");
 }
 
 void disarm(void)
 {
   _armed_state = DISARMED;
   LED1_OFF;
+  mavlink_log_warning("DISRMED");
```
dpkoch: Spelling! :)

src/mode.c
```diff
@@ -52,8 +56,8 @@ bool check_mode(uint32_t now)
       // if left stick is down and to the right
       if (pwmRead(_params.values[PARAM_RC_F_CHANNEL]) < _params.values[PARAM_RC_F_BOTTOM] +
           _params.values[PARAM_ARM_THRESHOLD]
-          && pwmRead(_params.values[PARAM_RC_Z_CHANNEL]) > (_params.values[PARAM_RC_Z_CENTER] + _params.values[PARAM_RC_Z_RANGE]/2)
-          - _params.values[PARAM_ARM_THRESHOLD])
+          && pwmRead(_params.values[PARAM_RC_Z_CHANNEL]) < (_params.values[PARAM_RC_Z_CENTER] - _params.values[PARAM_RC_Z_RANGE]/2)
```
dpkoch: I think we've already got an issue for this, but it'd be nice to have a separate RC module that handles all of the calibration stuff and just returns normalized values between -1 and 1 or 0 and 1 for each channel. But again, let's not worry about it here.

src/rc.c
```diff
@@ -30,29 +33,29 @@ static void convertPWMtoRad()
   // Get Roll control command out of RC
   if (_rc_control.x.type == ANGLE)
   {
-    _rc_control.x.value = ((pwmRead(_params.values[PARAM_RC_X_CHANNEL]) - _params.values[PARAM_RC_X_CENTER])
-                           *2*_params.values[PARAM_RC_MAX_ROLL_MRAD])/_params.values[PARAM_RC_X_RANGE];
+    _rc_control.x.value = (float)((pwmRead(_params.values[PARAM_RC_X_CHANNEL]) - _params.values[PARAM_RC_X_CENTER])
```
dpkoch: Again, just thinking that the RC module would only be responsible for returning normalized values, then the muxer would decide how to interpret them (apply the max roll/rollrate value based on the mode, etc.). Again, we can do that in a separate pull request

src/rc.c
```diff
@@ -30,29 +33,29 @@ static void convertPWMtoRad()
...
+                           *2.0f*get_param_float(PARAM_RC_MAX_ROLL_MRAD))/(float)_params.values[PARAM_RC_X_RANGE];
```
dpkoch: Probably just want to rename these params to get rid of the units now that they're in SI units (e.g. `PARAM_RC_MAX_ROLL` instead of `PARAM_RC_MAX_ROLL_MRAD`)

dpkoch:
It's sweet that this is working! I think that there is definitely some cleanup to do, but for the most part I'm good with merging as-is and cleaning up later. In addition to my  line comments above, it also seems like we've got a fair amount of code duplication with all the PID loops. Maybe it'd be possible to have a pid_t struct or something for each loop, then a generic PID function that runs an arbitrary loop on the values in that struct?

But yeah, maybe let's just take care of a couple of the minor fixes (spelling, parameter names) that could get lost easily, then worry about the more complicated changes in separate issues/pull requests?

superjax:
Sounds good to me!

I would totally be in favor of a pid_t or something like that.  I was trying to figure out a way to do it cleanly without a bunch of pointers, which is the only way I could think of to not have the code redundancy.  I figured that it would be easier to read in it's current form as opposed to a more complicated, but less redundant form.  You're better at that sort of coding than I am, so I am totally in favor of a refractor that uses less redundancy.

Let's make all of our improvements into issues and deal with them severally, rather than making some massive pull request that takes weeks to merge.

src/mavlink_receive.c
superjax: Sounds good.

src/mode.c
superjax: Check the rc_switch branch.  I'll do a pull request soon, (after this one)

src/rc.c
superjax: Well, we need to know what to control to in SI units. That's why I'm doing that in rc.c.  We want to be able to hit some setpoint in SI.

src/rc.c
superjax: I think I fixed this in my latest offboard branch.

superjax:
Sorry, I accidentally merged this.  master has been reverted, but since this pull request is closed, I'll have to open a new one. #65

PR #62, opened 2016-07-27
Offboard

superjax:
Few changes after #60, which incorporate RC trim setting and tested offboard control.

### DIFF
--- a/src/rc.c
+++ b/src/rc.c
@@ -6,12 +6,17 @@
 #include "param.h"
 #include "rc.h"
 #include "mux.h"
+#include "mode.h"
 
 #include "mavlink_util.h"
 #include "mavlink_log.h"
 
+bool _calibrate_rc;
+void calibrate_rc();
+
 void init_rc()
 {
+  _calibrate_rc = false;
   _rc_control.x.type = ANGLE;
   _rc_control.y.type = ANGLE;
   _rc_control.z.type = RATE;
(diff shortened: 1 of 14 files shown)


superjax:
By the way, I tested this by flying my AscTec quad with the joystick over ROS.  It worked great after incorporating RC trim settings.

src/rc.c
```diff
@@ -153,4 +162,66 @@ bool receive_rc(uint32_t now)
...
+          max[i] = read_value;
+        }
+        if(read_value < min[i])
+        {
+          min[i] = read_value;
+        }
+      }
+      delay(10);
+    }
+    _params.values[PARAM_RC_X_RANGE] = max[_params.values[PARAM_RC_X_CHANNEL]] - min[_params.values[PARAM_RC_X_CHANNEL]];
```
dpkoch: So I'm just wondering if storing the center and the range is sufficient for all cases? For example, if the trim value isn't exactly centered between the endpoints, how do we interpret the values? Would it be better to store min, max, and trim values so that we can completely characterize the channel, or is this sufficient in your opinion?

Also, should we have a direction param for each channel so we can flip channels if necessary?

superjax: I think that we can require that the range is equally spaced about the center.  (It would be pretty weird to have something else).

I have a pull request pending (rc_switch branch) that builds a switch_t with a direction.  As for the actual channels, the transmitter is a more convenient place to have the direction switch.  (I used to have a direction switch for each, but then I messed it up so many times that I just decided to leave it on the transmitter, rather than dealing with it in code.)

Issue #69, opened 2016-08-02
Convert mixer to float


superjax:
We want to switch the mixer to be in float, rather than integer-based math.  This will require re-tuning the PID loops.

Issue #68, opened 2016-08-02
Integrator Anti-Windup


superjax:
We just need to add and integrator anti-windup scheme.  (Perhaps Section 6.5 of the UAV book?)

Issue #65, opened 2016-08-02
Reduce Code Redundancy in controller.cpp


superjax:
Per my discussion with @dpkoch in #60, Maybe it'd be possible to have a pid_t struct or something for each loop, then a generic PID function that runs an arbitrary loop on the values in that struct.

PR #72, opened 2016-08-17
Controller struct

superjax:
This still needs to be tested on hardware, but It solves #69, #65, and #68.

@dpkoch, do you want to look at my refractoring of controller.c, and let me know if this is the sort of thing you were thinking about?


dpkoch:
Sweet! Is this still a WIP? (it's good to remove that from the name once you think it's ready.) If so cool, but if you're done then I have a couple of edits to make before we merge.

superjax:
It's still WIP until I test it on hardware.  I just haven't been near a quad for a while.

PR #79, opened 2016-09-15
Async

superjax:
There are some issues with initialization that I need to work out still, but I dropped the `update_sensors()` to 27 us.  That's 1/20 the time it used to take.

It's a pretty big merge, so I thought I would push the working code up right now, while I'm still working on it so you can start going over it.


superjax:
Alright @dpkoch.  I think this is about ready.  At least, it's enough of the way that it can be the master.

In this branch, magnetometer, sonar, IMU, and diff pressure are working.  Barometer has issues.  Every once and a while, it seizes up.  I've finally bought a debugger, which I intend on using to help me figure out why it stops responding.  I think that bug will have to be solved in another pull request.

PR #82, opened 2016-09-19
Param API

dpkoch:
Converting everything over to using parameter get and set methods in preparation for making `_params` struct private. Still need to factor flash.c so that it doesn't use the `_params` struct, but everything else should be good to go. I'm thinking we merge in this much in the meantime. Just need to test functionality to make sure everything is still working correctly before we merge.

### DIFF
--- a/include/controller.h
+++ b/include/controller.h
@@ -12,9 +12,9 @@ extern "C" {
 
 typedef struct
 {
-  float* kp;
-  float* ki;
-  float* kd;
+  param_id_t kp_param_id;
+  param_id_t ki_param_id;
+  param_id_t kd_param_id;
 
   float* current_x;
   float* current_xdot;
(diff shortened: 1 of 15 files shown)


dpkoch:
@superjax Got that white rev5 Naze32 up and running on the ROScopter quad and tested it out. I just used the gains that were saved in the `params.yaml` file on the odroid. The control felt maybe a little soft and the yaw drifted a bit, but other than that it flew pretty well. Is that consistent with the type of performance you were getting before? If so then I think this is ready to merge. The fact that I was able to load the params was a good sign too.

superjax:
Awesome!  

I never put the tuned gains into the `param.c`.  I kept programming them in each time, because I was kinda lazy.

Awesome!  Thanks for getting that working.

Issue #28, opened 2016-06-20
Arming Logic


superjax:
We should re-code mode.c to be a state machine, rather than a series of if and else statements

superjax:
I looked into doing this, and I think that it would be more confusing that the if statements that we have.  There really are only four states.

Anyway, I think that it's fine as is.  I'm wondering if @dpkoch had insight that I don't have.

dpkoch:
#72 has a hotfix for issue where autopilot would arm on initialization if RC was not present. This issue should address this in a more sophisticated fashion.

superjax:
A lot of cleanup happened in #103 with the RC abstraction.  I think that we might still get more clarity from a state machine matrix approach, but until then, I think this is pretty good.

PR #90, opened 2017-01-26
Release Candidate 1.0

superjax:
This is the pull request for release candidate 1.0


src/estimator.c
```diff
@@ -27,75 +27,79 @@ static const vector_t g = {0.0f, 0.0f, -1.0f};
...
-
-  w_acc.x = 0.0f;
-  w_acc.y = 0.0f;
-  w_acc.z = 0.0f;
-
-  q_tilde.w = 1.0f;
-  q_tilde.x = 0.0f;
-  q_tilde.y = 0.0f;
-  q_tilde.z = 0.0f;
+    _current_state.q.w = 1.0f;
```
superjax: I added a reset_state() function so that when we perform an IMU calibration, the state doesn't jump and potentially go unstable.

src/mode.c
```diff
@@ -20,10 +21,23 @@ void init_mode(void)
...
 
-void arm(void)
+bool arm(void)
 {
-  _armed_state = ARMED;
-  LED1_ON;
+  static bool started_gyro_calibration = false;
+  if (!started_gyro_calibration && _armed_state == DISARMED)
+  {
+    start_gyro_calibration();
```
superjax: This is hacky.  I think once we move to a better state machine then this will work better.  For now, I just wanted to run a gyro calibration on arm, so the arm() function returns a boolean of whether or not it succeeded.  The logic works fine for now, but I am certain that it will need to change in the future.

src/param.c
```diff
@@ -208,13 +237,27 @@ void param_change_callback(param_id_t id)
...
+    break;
 
   case PARAM_STREAM_SERVO_OUTPUT_RAW_RATE:
     mavlink_stream_set_rate(MAVLINK_STREAM_ID_SERVO_OUTPUT_RAW, get_param_int(PARAM_STREAM_SERVO_OUTPUT_RAW_RATE));
     break;
   case PARAM_STREAM_RC_RAW_RATE:
     mavlink_stream_set_rate(MAVLINK_STREAM_ID_RC_RAW, get_param_int(PARAM_STREAM_RC_RAW_RATE));
     break;
+
+  case PARAM_RC_TYPE:
```
superjax: Adding a few callbacks here makes these parameters hot-swappable.

src/main.c
```diff
@@ -28,46 +28,26 @@ int main(void)
   SetSysClock(0);
   systemInit();
 
-  // Perform Setup Operations
-  setup();
-
-  // Initialize Serial ports
-  Serial1 = uartOpen(USART1, NULL, get_param_int(PARAM_BAUD_RATE), MODE_RXTX);
```
dpkoch: Can we move the `serialPort_t *Serial1;` (line 25) to `mavlink.c` as well?

src/mode.c
dpkoch: Cool, sounds good. Yeah we can clean this up when do a better state machine. I guess we already have an issue for improving the state machine (#28) so this should be captured by that, right?

src/param.c
```diff
@@ -33,6 +33,10 @@ static void init_param_float(param_id_t id, char name[PARAMS_NAME_LENGTH], float
 // function definitions
 void init_params(void)
 {
+  for(uint8_t i = 0; i < PARAMS_COUNT; i++)
+  {
+    init_param_int(i, "DEFAULT", 0);
+  }
```
dpkoch: I used to have something like this here because we weren't initializing every param before, but once that was fixed I got rid of it because it seems like a lot of unnecessary work. Is there a specific reason you put it back in?

src/param.c
```diff
@@ -46,129 +50,154 @@ void init_params(void)
...
+  init_param_int(PARAM_ELEVATOR_REVERSE, "ELEVATOR_REV", 0); // reverses elevator servo output | 0 | 1
+  init_param_int(PARAM_AILERON_REVERSE, "AIL_REV", 0); // reverses aileron servo output | 0 | 1
+  init_param_int(PARAM_RUDDER_REVERSE, "RUDDER_REV", 0); // reverses rudder servo output | 0 | 1
+
+  /********************/
+  /*** ARMING SETUP ***/
+  /********************/
+  init_param_int(PARAM_ARM_STICKS, "ARM_STICKS", true); // use RC sticks to arm vehicle (disables arm RC switch if enabled) | 0 | 1
+  init_param_int(PARAM_ARM_CHANNEL, "ARM_CHANNEL", 5); // RC switch mapped to arm/disarm [0 -indexed] | 4 | 7
+  init_param_int(PARAM_ARM_THRESHOLD, "ARM_THRESHOLD", 150); // RC deviation from max/min in yaw and throttle for arming and disarming check (us) | 0 | 500
```
dpkoch: Nice!! This is awesome

src/param.c
dpkoch: Cool. I've been thinking about the param callback structure; I think I'd like to change it someday so that each module just has a callback method, then if params change we just call them all and the individual modules are responsible for deciding what to do about it. Some day...

src/sensors.c
```diff
@@ -11,63 +11,198 @@ extern "C" {
...
 
-static bool calib_gyro;
-static bool calib_acc;
 
+bool update_sensors()
+{
+  // Look for disabled sensors while disarmed (poll every 10 seconds)
+  // These sensors need power to respond, so they might not have been
+  // detected on startup, but will be detected whenever power is applied
+  // to the 5V rail.
```
dpkoch: Cool. Do you think we should up that rate a bit? This shouldn't take all that much extra time

src/sensors.c
```diff
@@ -11,63 +11,198 @@ extern "C" {
...
+  static uint32_t last_time_look_for_disarmed_sensors = 0;
+  if (_armed_state == DISARMED && (!_sonar_present ))//|| !_diff_pressure_present))
```
dpkoch: I think this logic wouldn't work if a sonar was plugged in but not a diff pressure sensor...? But it's also redundant with the checks in the inner if block; maybe only check for `_armed_state == DISARMED` here?

src/sensors.c
```diff
@@ -76,145 +211,157 @@ static bool update_imu(void)
...
   {
+    // if we have lost 1000 IMU messages something is wrong
+    if(millis() > last_imu_update_ms + 1000)
+    {
+      // change board revision and reset IMU
+      last_imu_update_ms = millis();
+      set_param_int(PARAM_BOARD_REVISION, (get_param_int(PARAM_BOARD_REVISION) >= 4) ? 2 : 5);
+      uint16_t acc1G;
+      mpu6050_init(true, &acc1G, &gyro_scale, get_param_int(PARAM_BOARD_REVISION));
+      accel_scale = 9.80665f/acc1G * get_param_float(PARAM_ACCEL_SCALE);
```
dpkoch: @superjax What's the purpose/function of the `ACCEL_SCALE` parameter?

include/sensors.h
```diff
@@ -12,27 +12,30 @@ extern "C" {
 extern vector_t _accel;
 extern vector_t _gyro;
 extern float _imu_temperature;
-extern uint32_t _imu_time;
-extern bool _imu_ready;
+extern uint64_t _imu_time;
 
 extern bool _diff_pressure_present;
```
dpkoch: Or rename the data fields to `_diff_pressure`, `_diff_pressure_velocity`, `_diff_pressure_temp`? That would probably be cleaner

src/sensors.c
superjax: The ACCEL_SCALE parameter takes care of the weird MPU6050's that are sometimes on the naze32's. You know, that one that you had for a long time that for some reason reports half of full range.  This routine takes care of that.

I'm not sure how often that happens.  It might be that we could just ignore it, and assume that everyone has a good IMU.

src/sensors.c
superjax: You're right.  I'll fix that.

src/sensors.c
superjax: I don't know, I feel like twice a second is plenty fast.  I actually feel like once every few seconds would be sufficient.  How fast are you thinking?

I2C is the slowest thing we do, so I'm just trying to keep it to a minimum.

superjax: Haha, upon reading this, I realize that my note does not agree with the code.   Perhaps I'll adjust the comment.

src/param.c
superjax: Did you see my markdown [auto generator](https://gist.github.com/superjax/4e23bda4908d4cf13d0ade4939de6165)? 

It parses the param.c file and creates markdown for the wiki.  You'll notice that it's been updated recently!

src/param.c
superjax: Yeah, it's because it seems that if the parameters are not initialized, then they don't get sent to fcu_io.  Instead, fcu_io just keeps requesting parameters because it doesn't get them all.

I put that in there mainly as a way to ensure that fcu_io got all the paramters, and so I could see which parameters weren't being initialized.  I think now that I have made sure that param.h and param.c agree on the number of parameters we can get rid of this.

src/mode.c
superjax: Yeah, I imagine that a new and improved state machine would take care of that.

src/main.c
superjax: Yeah. that's a good idea.

src/flash.c
```diff
@@ -73,11 +73,17 @@ bool writeEEPROM(bool blink)
...
+    for(int i = 0; i < NUM_PAGES; i++)
+      status = FLASH_ErasePage(FLASH_WRITE_ADDR + i * FLASH_PAGE_SIZE);
     for (unsigned int i = 0; i < sizeof(params_t) && status == FLASH_COMPLETE; i += 4)
+    {
       status = FLASH_ProgramWord(FLASH_WRITE_ADDR + i, *(uint32_t *)((char *)&_params + i));
-    if (status == FLASH_COMPLETE)
+      if (status != FLASH_COMPLETE)
+      {
+        continue;
+      }
```
superjax: Yeah, I don't really know why I did that. You're probably right.  I'll figure out what is going on here and do it right.

include/sensors.h
superjax: Good idea!

include/estimator.h
```diff
@@ -10,20 +10,29 @@ extern "C" {
...
 typedef struct
 {
-  float p;
-  float q;
-  float r;
-  float phi;
-  float theta;
-  float psi;
+  quaternion_t q;
+  vector_t euler;
```
superjax: That's a good idea.

src/flash.c
superjax: So I think this is the correct behavior.  If that statement returns true, then later on in the code, we return false.

src/param.c
superjax: I did this on the CXX branch with std::bind and std::function.  It's pretty sweet!

superjax:
This is the biggest pull request anyone ever did.  Let's not do it this way again.

PR #103, opened 2017-02-27
[WIP] Board support layer

dpkoch:
I decided to finally just tackle the board support layer stuff; I've got some ideas for how to do SIL with ROS that I want to try out. There are some things that could probably be cleaned up, but for now I'm just keeping it as close as possible and leaving some of the cleanup for later.

I've still got a few things to finish up before this is done (convert the system clock stuff over and modify the make file), but I thought I'd let you know what I'm up to. I'll try to finish that tonight.

I'm not sure if you want to merge this in before RC1.0, but if you like it then I would also like to not let it get too far out sync before merging since I'd probably have to bring some updates across manually. Either way, let me know what you think.


superjax:
This is awesome!! I"ll look through it tonight.  @drinkdhmo has experience with the STM32 HAL, so I'm making sure he is a part of this conversation too, since he was planning to be involved in the F4 bringup.

plusk01:
Yeah, this is great @dpkoch -- it's perfect timing for a project that I'm working on where I'm considering running ROSflight on the [BeeCore F3](http://www.banggood.com/Beecore-F3_EVO_Brushed-ACRO-Flight-Control-Board-DSM2-For-Inductrix-Tiny-Whoop-Eachine-E010-p-1089021.html).

src/main.c
```diff
@@ -18,15 +19,9 @@
 #include "mixer.h"
 #include "rc.h"
 
-extern void SetSysClock(bool overclock);
-
-serialPort_t *Serial1;
-
 int main(void)
```
superjax: It would be super nice to bring back the `init()` and `loop()` functions, so we can use them in the SIL (can't have two mains, so I can't call main() directly, and there is a bunch of stuff I have to do in between loops). I don't like being arduino-y though, because it makes me feel like a goof.  What do you think?

src/mavlink_receive.c
```diff
@@ -212,10 +213,9 @@ static void handle_mavlink_message(void)
 // function definitions
 void mavlink_receive(void)
 {
-  while (serialTotalBytesWaiting(Serial1))
+  while (serial_bytes_available())
```
superjax: Do we want to every support more than one UART?  For example, both a USB connection and a serial radio?  Or does that get into feature creep?

include/board.h
```diff
@@ -0,0 +1,58 @@
...
+
+// setup
+void init_board(void);
+
+// clock
+uint32_t millis();
+uint64_t micros();
+
+// serial
+void serial_init(uint32_t baud_rate);
```
dpkoch: Agreed. I'm thinking we leave it as is for now and then address that later?

include/board.h
```diff
@@ -0,0 +1,58 @@
...
+void serial_write(uint8_t byte);
+uint16_t serial_bytes_available(void);
+uint8_t serial_read(void);
+
+// sensors
+void sensors_init(int board_revision);
+
+void imu_register_callback(void (*callback)(void));
+void imu_read_accel(float accel[3]);
```
dpkoch: Yeah I wasn't sure about that... It might eliminate a couple of unnecessary copies. Maybe I'll reevaluate once I get the build stuff figured out

src/main.c
dpkoch: Haha, yeah. I was thinking we could maybe just have like a `main_loop` or `run` function or something that just lumps it all together; do you think there's value in having it split up? We could call it `init()` and `run()` maybe so it doesn't sound quite so arduino-y

src/mavlink_receive.c
dpkoch: I think that could potentially be good; it'd let us support DSM receivers and I don't think it'd be too much feature creep. Of course, that could be handled on the board support side too maybe and be transparent to the flight stack. I think this is something we can leave as-is for now, but will probably want to revisit. The whole RC interface could probably use some cleaning up in general anyway

dpkoch:
Okay, well, it builds now. There's one warning I still need to track down, but it builds. I haven't tested anything yet though, I still don't have a Naze for some reason...

superjax:
So I just caught this up with the most recent RC1.0.  There are still some things to take care of (Like, it won't calibrate IMU anymore, and the magnetometer detection is broken, the Makefile is a little messed up), but it builds now!

superjax:
I think everything is about ready.  Tomorrow morning, I'll make sure to fly it on the multirotor and we'll be good to go!

dpkoch:
Was just playing with this (finally got a Naze! Flip32+ rev2.5), and the parameter write seems to be broken. fcu_io reports success, but the green LED never flashes and changes do not persist across reboot. Don't have time to dig into right now, but figured I'd at least report

superjax:
Have you tried this with the commits I made earlier this evening?  commit e7f2de2 should have fixed that issue.

sytelus:
Here are the similar changes I'd been doing: https://github.com/byu-magicc/ROSflight/issues/110

dpkoch:
@superjax Yes, I just double checked, but I'm having this issue even with that commit. :(

superjax:
That is weird.  I have independently tried this on 2 separate nazes and cannot replicate it. it seems to be working as expected.

dpkoch:
Strange. Which boards/versions have you tried it on? Mine's a Flip32+ rev2.5 from RTF quads.

PR #118, opened 2017-03-17
Make offboard control compatible with float mixer, fix mixer servo wr…

dpkoch:
Fixed what I think are three bugs:
  1. `mavlink_receive` was still converting offboard control messages to NWU, but everything is in NED now, right?
  2. `mavlink_receive` was putting pass-through values in units of PWM, but it should just be floats now, right?
  3. I think the scaling on servo write was off (would swing from 500 to 2500 instead of 1000 to 2000)


superjax:
I agree.  Thank you!

Issue #117, opened 2017-03-17
Remove altitude hold mode?


dpkoch:
@superjax As I've been working on cleaning up the mux stuff, I'm starting to think we should remove the altitude hold stuff for now. I think it would be a nice feature, but (a) it's already bordering on "feature creep" anyway, (b) it really makes things a lot messier, and (c) the fact that we still haven't implemented it makes me worry it would hold up our 1.0 release too much if we're still going to promise it being in there.

So, what if we remove it entirely from the 1.0 release, then we can decide later if we want to put it back in? We can always get altitude control in the meantime by writing a controller on the ROS side that sends only throttle commands, which may just be a better long-term solution anyway. Let me know what you think.

superjax:
I would completely support that decision.  I wasn't entirely sold on the the idea to begin with, and my attempts at getting a Lyuapunov-stable altitude observer have not proven successful.

superjax:
We removed this as part of #103

PR #120, opened 2017-03-17
[WIP] RC/mux cleanup

dpkoch:
This is maybe more extensive than what you were hoping for, but I think it's a step in the right direction. One of my main goals was to improve the encapsulation between the different modules. I think I pretty much accomplished that between RC and mux. With this change the RC module is responsible only for producing normalized stick values (-1 to 1 or 0 to 1) and switch values (boolean). The mux module does all interpretation of what the sticks mean. There's still a little bit of a goofy dependency between mavlink and the mux module, but I'm not 100% percent sure how to resolve without one module needing to know too much about the internals of the other.

I also made a few changes to the muxing:
  - Added attitude mode and throttle mode parameters, so you can still set that even if you don't have a switch set (which is the default now, I figure we don't want a switch doing that unless the user actually knows it's there and wants it)
  - Added the ability to not have a switch set for the different actions
  - The muxer now checks for stick deviation overrides in each direction independently; that way if you're drifting into a wall you can just give a little bump the side without affecting the other channels. I believe that's the way our PX4 setup was doing it
  - Everything external to the RC module sees RC as normalized; all scaling and calibration happens internally and is invisible to everyone else. This also means that we're really close to being able to have the board support layer provide normalized values. We'd just have to update our calibration stuff to be in floats, but I think it actually wouldn't change all that much.

I also tried to clean up the data structures for the muxing stuff, but haven't quite figured that out yet. However, this new logic flow reduces the dependency on the structure a little I think, so it will hopefully be easier to clean up in the future.

I haven't tested this yet; I'll try to get to that tomorrow.

Logic diagram for interpreting RC:
https://drive.google.com/file/d/0BxQ-B33JcLpgZVdLQkxYTGtSZzQ/view?usp=sharing

Logic diagram for muxing:
https://drive.google.com/file/d/0BxQ-B33JcLpgVjBHeG1GYjRfQUE/view?usp=sharing


superjax:
I like all of it.  Thank you!

I went ahead and removed altitude mode per #117, and cleaned up the mixer to go with it. This is a really nice change.

superjax:
I'm going to go ahead and merge this into the `board` branch right now, so I can go test it tomorrow.  There isn't anything super special about the board branch right now, except that it has sort of become a scratch pad.  If it breaks, no big deal.

Issue #122, opened 2017-03-19
Documentation


superjax:
Hey @dpkoch, @el3ment, @dallinbriggs, @plusk01, @jerelbn. Thanks for interest in this project.  I am getting excited about an official release!  

As I've been thinking about RC1.0 (and helping people get going in the lab on ROSflight),  I am wondering if we should also release some real documentation.  My primary concern is the "user documentation" (rosflight.org, and the github wiki), not so much the doxygen "code documentation," which is more straight-forward, and less useful to newbies than a user guide.

I'm thinking the user guide (in whatever form it finally takes) could have the following outline.  I expect our  main audience to be people like the USU grad students, people in our lab, or people like Insitu, or AFRL who want to use our stuff.  I've got an example of what I'm thinking below.

My question to you is: 
1. Is this too much?  Too little?  Is there something I'm forgetting or don't need?
2. Would you be willing to help generate some of this?
3. Do you have ideas of how to present the documentation?  [ROSflight.org](http://www.rosflight.org) is really not that great right now.  It's super disorganized.  I suspect that there is a better tool for user guides online.

Anyway.  Thanks!

- Getting Started
- Introduction
    + Purpose
    + Parts List
    + Best Practices
- Installing and running Software
    + Installing software (ros, eigen, 32-bit dependencies, stm32flash etc...)
    + Description of `fcu_*` and `ros_*` packages
    + Flashing ROSflight
    + Running fcu_io
    + Base Station Setup? (This is mostly ROS networking, maybe we can punt to the ROS tutorials?)
- Putting the Hardware Together
    + Wiring Diagram
    + Motor Layouts
    + External Sensors (GPS, Sonar, Airspeed)
- Preflight Checks
    + Common Parameters to check
    + Checking control surface deflections/motor directions
    + Perfoming Sensor Calibration
- Getting good performance
    + Multirotor Gain Tuning guide
    + RC Calibration and trim calculation
- Autonomous Flight
    + Provide control from onboard computer
    + Fly waypoints with `ros_copter`
    + Fly waypoints with `ros_plane`

- Advanced - These may or may not be important to RC1.0
    + Custom Mixers
    + Temperature Calibration
    + Saving parameter files and sharing them to other aircraft
    + Extending `ros_copter` and `ros_plane` example classes
    + Simulating ROSflight with `fcu_sim`
    + Developing custom firmware code
    + Extending ROSflight to other flight controllers (board layer API)

[...]

PR #126, opened 2017-03-23
Documentation

dpkoch:
Here's what I'm thinking for the documentation. I decided to just put it on readthedocs for now; we can always move it later if we want to. You can see a live demo at http://docs.rosflight.org/en/documentation/ (once it's all merged in you'll be able to get there straight from docs.rosflight.org).

I also tweaked the outline a bit and posted it in #122. Let me know what you think.


superjax:
Sorry about the build error.  That was some half-finished work I accidentally pushed.  That's all on another branch I'm working on.

superjax:
Hey guys, I have a draft for the core of the documentation.  I would really love it if someone would go through and make some suggestions to help improve what I wrote.

dpkoch:
@superjax I've finished all the pages now. They may not be perfect, but unless you catch any glaring errors I say we merge them in and call it good enough for release. The readthedocs page will by default show whatever's on the latest master anyway, so we can continue editing the documentation even between releases.

superjax:
I like it!

Issue #124, opened 2017-03-23
IMU Calibration


superjax:
@dpkoch, @plusk01, @jerelbn Do we want to improve the IMU calibration routine for RC1.0?

jerelbn:
I say yes and I will work on it.

superjax:
This was partially closed by #131

superjax:
We are moving this out of the firmware and into the ROS stack

PR #129, opened 2017-03-31
Imu fix

superjax:
This fixes some bugs I found in the IMU calibration routine.  This does not completely address #124, but it is a start.


superjax:
I just flew this successfully on hardware.

superjax:
Merged into #131

Issue #128, opened 2017-03-30
Failsafe support


jerelbn:
We need to be able to modify hover throttle command.

plusk01:
It doesn't look like the `check_failsafe` logic in [mode.c](https://github.com/byu-magicc/ROSflight/blame/status/src/mode.c#L68) is right... 

```c
bool check_failsafe(void)
{
  if (pwm_lost())
  {
    _armed_state |= FAILSAFE;
  }

  else
  {
    for (int8_t i = 0; i<get_param_int(PARAM_RC_NUM_CHANNELS); i++)
    {
      if(pwm_read(i) < 900 || pwm_read(i) > 2100)
      {
        _armed_state |= FAILSAFE;
      }
    }

    _armed_state &= ~(FAILSAFE); // <-- This will always cause an out-of-range 
    return false;                // <-- signal failsafe to be ignored
  }

  if (_armed_state & FAILSAFE)
  {
    static uint8_t count = 0;
    if (count > 25)
    {
      led1_toggle();
      count = 0;
    }
    count++;
    return true;
  }
}
```

Also, it looks like we have three failsafe branches -- what is the state of everything?

superjax:
Thanks @plusk01.

You're right.  That logic is dumb.  Thanks for catching that.  We should update it on the `failsafe_support` branch and incorporate it into #131

The `failsafe_support` branch supports setting the hover throttle.  It should be ready to merge (besides this bug)

We are trying to decide how to handle #132, which would require offboard computer commands to be accepted even if we are in failsafe.  The work on that is on the `offboard_failsafe` branch and is still a work in progress.  Don't worry about that branch for now.

superjax:
@plusk01, do you any ideas on how to fix this?

plusk01:
How about

```c
bool check_failsafe(void)
{

  bool failsafe = false;

  if (pwm_lost())
  {
    // Set the FAILSAFE bit
    failsafe = true;
  }
  else
  {
    // go into failsafe if we get an invalid RC command for any channel
    for (int8_t i = 0; i<get_param_int(PARAM_RC_NUM_CHANNELS); i++)
    {
      if(pwm_read(i) < 900 || pwm_read(i) > 2100)
      {
        failsafe = true;
      }
    }
  }

  if (failsafe)
  {
    // blink LED to let the user know we are in failsafe
    static uint8_t count = 0;
    if (count > 25)
    {
      led1_toggle();
      count = 0;
    }
    count++;

[...]

Issue #125, opened 2017-03-23
Static Offset Calibration


superjax:
I don't love the way that equilibrium forces and torques are applied.  I want to fix it.

superjax:
I have a possible solution for this on the `force_offsets` branch

superjax:
Closed by #131

Issue #121, opened 2017-03-17
offboard control timeout


superjax:
We should probably revert to RC or failsafe if we stop receiving offboard command messages and expect them.

superjax:
Closed by #131

PR #131, opened 2017-03-31
Failsafe command integration

superjax:
I had to squash the three pull requests into one in order to test on hardware.

Addresses #128 and #121, part of #124 and #125


superjax:
This is tricky to test.  @jerelbn and I will try to fly this this week.

superjax:
We tested this in hardware.  There is still no way for me to tell if the RC link has been lost (unless failsafe is configured on the rc reciever), but failsafe is working.

To get failsafe working, I had to squash all three of these pull reqeusts into one.  They are all working simultaneously.  The IMU fix, the force offsets, and the failsafe.

Closes #128, #125, and part of #124

src/controller.c
```diff
@@ -242,15 +242,54 @@ void run_controller()
...
+    set_param_float(PARAM_Z_EQ_TORQUE, _command.z);
+
+    mavlink_log_warning("Equilibrium torques found and applied.");
+    mavlink_log_warning("Please zero out trims on your transmitter");
+  }
+  else
+  {
+    mavlink_log_warning("Cannot perform equilbirum offset calibration while armed");
+  }
+}
```
dpkoch: This is pretty cool, it should work even if RC is on angle mode and onboard is in rate mode or vice versa, right? That's pretty slick. A couple of potential problems I see:
 - Probably need to reset those torque params to zero before doing this calibration, since otherwise the old values would get applied and mess up your new values
 - This assumes that you land on a level surface. If you were flying in the park and landed at an angle in the grass, then ran this command, it would mess up your trim values (if you were tilted in the direction of your trim, the error between desired and actual would be zero so your control will be zero, while if you're tilted the opposite way, the computed torque would be extra big). Could we maybe temporarily overwrite `_current_state` to zero for this step to get around that or something?

superjax: That's a great idea.   Thanks for catching that.

src/mixer.c
```diff
@@ -99,7 +99,14 @@ static mixer_t *array_of_mixers[NUM_MIXERS] =
 
 void init_mixing()
 {
-  // We need a better way to choosing the mixer
+  uint8_t mixer_choice = get_param_int(PARAM_MIXER);
+
+  if (mixer_choice >= NUM_MIXERS)
+  {
+    // we should probably enter an error state here
```
superjax: Let's issue an error message here.

src/sensors.c
```diff
@@ -184,9 +186,14 @@ static bool update_imu(void)
   if (new_imu_data)
   {
     last_imu_update_ms = clock_millis();
-    imu_read_accel(accel);
-    imu_read_gyro(gyro);
-    _imu_temperature = imu_read_temperature();
+    imu_read_all(accel, gyro, &_imu_temperature);
+
+    // Sometimes the IMU returns all zeros (this is a bug in the STM32F1 I2C driver)
```
superjax: Let's move this to the board support layer

PR #134, opened 2017-04-05
Saturation fix

### LINKED ISSUE #127: PARAM_MAX_COMMAND not converted to float
It looks like PARAM_MAX_COMMAND got missed when everything got converted over to float; it's currently an int with default value of 1000. So it's probably doing nothing. I assume it should be a float with default value of 1.0? @superjax if you agree then I'm happy to make the switch.

superjax:
Fixes #127


superjax:
Flew on Hardware, works

Issue #132, opened 2017-03-31
Failsafe auto flight termination


mmmfarrell:
For fixed wing ideally we would like flight termination to be: throttle closed, full up elevator,
full right rudder, full right or left aileron. 

It would be nice to have different levels of fail safe, i.e. after 30 seconds of communications loss, the aircraft automatically Returns to Home or Returns To Land. After 3 minutes of communication loss, the aircraft terminates flight. 

Also would be ideal if RTH/RTL and flight termination were activatable by the safety pilot and ground station. 

(These are our requirements for AUVSI)

superjax:
I think the cleanest way to do this is to publish a `status` message from fcu_io.  This could include things like whether or not the safety pilot is in control, we are in failsafe, loop time, armed/disarmed, etc...  I think that the RTH/RTL and flight termination should probably occur within `ros_plane`.

The other change necessary is that we need to allow offboard control to be processed even if RC is lost.

@dpkoch, do you have any thoughts?

superjax:
I just added a status message which is published by `fcu_io`.  This includes a flag of whether or not the flight controller is in failsafe mode.  You'll have to watch this flag and act appropriately to fulfill your requirements.

As of right now, we default to RC.  However, we could allow onboard computer commands to be executed even in the absence of RC.

This should be supported by the `status` branches of `fcu_io` and `fcu_common` and the `status_message` branch of `ROSflight`.  That is, until we merge into `RC1.0`

superjax:
The latest `status_message` branch should allow for onboard commands to be accepted even if RC is lost.

mmmfarrell:
awesome. I will work on handling this message in ROSPlane

superjax:
When you get around to testing this, I can show you a quick hack that tricks ROSflight into thinking it's lost RC when you flip a switch.

superjax:
We should just make this a parameter so it can be turned on, but by default is turned off.

bsutherland333:
Failsafe behavior is something we'd like to revisit, so I'll leave this open.

PR #135, opened 2017-04-07
[WIP] - Status message

superjax:
This follows #131.

This includes a status message per #132, and it also allows onboard commands to be accepted in the event of RC loss (for RTH functionality).  Still needs to be tested.

@dpkoch, I'm sorry, I had to mess up your beautiful `mux.c` functions.  If you have a better way to do this logic, then I'm all ears.  We now have 3 commands we have to mux, instead of 2.  Anyway.  I'd love your input.


src/mavlink_stream.c
```diff
@@ -30,21 +67,14 @@ static void mavlink_send_heartbeat(void)
...
-  }
-  else
-  {
     if(rc_switch(get_param_int(PARAM_RC_ATT_CONTROL_TYPE_CHANNEL)))
       control_mode = MODE_ROLL_PITCH_YAWRATE_THROTTLE;
     else
       control_mode = MODE_ROLLRATE_PITCHRATE_YAWRATE_THROTTLE;
-  }
 
   mavlink_msg_heartbeat_send(MAVLINK_COMM_0,
```
superjax: I don't love how much I'm abusing this message.  I considered creating a ROSFLIGHT_STATUS message which could be a lot easier to understand (as opposed to hijacking the default MAVLink stuff.)  What do you think?

superjax:
Flew on hardware, seems to be working properly.

src/mux.c
```diff
@@ -137,60 +158,77 @@ static bool stick_deviated(mux_channel_t channel)
 
 static bool do_roll_pitch_yaw_muxing(mux_channel_t channel)
 {
-  bool rc_override;
+  bool rc_override = (rc_switch_mapped(RC_SWITCH_ATT_OVERRIDE) && rc_switch(RC_SWITCH_ATT_OVERRIDE))
+                      || stick_deviated(channel) || muxes[channel].rc->active;
```
dpkoch: Won't this always go into rc override if not in failsafe mode since rc->active is always true if not in failsafe mode and you've got an OR here? I was really trying to move way from these active fields since in my opinion they make the logic more confusing; the only reason I didn't delete it entirely was because I needed some way of knowing which values were valid in the offboard control message.

superjax:
I think I fixed the logic.

So, the reason I implemented this is because of #132.  I believe that there is a solid case for RTH on rc dropout.  Not completely sold, but I think that is the use case.

PR #139, opened 2017-04-21
[WIP] - Offboard failsafe

superjax:
This PR incorporates offboard commands in the event of RC dropout.  This is so we can support return-to-home functionality as proposed in #132.


superjax:
I'm going to close this for now, because we are not supporting it in RC1.0

Issue #147, opened 2017-05-03
[Minor] Incorrect indices used in init_switches


len0rd:
The switch statement within rc.init_switches() uses incorrect indices, resulting in none of the cases running. The 'chan' iterator goes from 0-3, which is what should be used for the cases.

superjax:
@len0rd, could you make a pull request for this issue?

PR #150, opened 2017-05-04
Fix for issue #147:

len0rd:
Fix for issue #147:Incorrect case statements on switch within rc:init_switches


superjax:
We should probably squash all these when we merge (I tried rebasing and it made a mess of the git history)

Issue #151, opened 2017-05-05
Fixes / Code Cleanup for controller


len0rd:
As discussed yesterday, there are some small fixes and optimizations we can apply to  controller.run_pid

PR #152, opened 2017-05-05
Fix for issue #151: removed unnecessary or buggy else statements. Now…

len0rd:
… uses _combined_control to check if throttle is high


superjax:
Thanks @len0rd

Issue #143, opened 2017-05-02
Version Info


superjax:
#defined by make, published in status message

superjax:
An idea of how to do this is on the `git_version` branch.  (I baked it into the Makefile)

dpkoch:
Closed by #156

PR #156, opened 2017-05-08
Automatic git version control baked into code.

superjax:
I don't know quite what to do if the git is dirty.  Should we XOR the hash if it's dirty?

Also, do we want to add this to the status message or heartbeat message?

### DIFF
--- a/naze/Makefile
+++ b/naze/Makefile
@@ -170,12 +170,24 @@ OPTIMIZE = -Os
 LTO_FLAGS = -flto -fuse-linker-plugin $(OPTIMIZE)
 endif
 
+#################################
+# VERSION CONTROL
+#################################
+GIT_VERSION := $(addprefix 0x,$(shell git describe --abbrev=4 --always))
+GIT_DIRTY := 0
+ifneq ($(shell git status --porcelain), )
+GIT_DIRTY := 1
+endif
+GIT_VARS:=$(addprefix -DGIT_VERSION=, $(GIT_VERSION)) $(addprefix -DGIT_DIRTY=, $(GIT_DIRTY))
+
+
 
 #################################
 # Flags
 #################################
 MCFLAGS=-mcpu=cortex-m3 -mthumb
-DEFS=-DTARGET_STM32F10X_MD -D__CORTEX_M3 -DWORDS_STACK_SIZE=200 -DSTM32F10X_MD -DUSE_STDPERIPH_DRIVER
+$(info GIT_VARS=$(GIT_VARS))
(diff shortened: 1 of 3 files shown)


plusk01:
I'm not sure what you mean by XOR'ing the hash (like, do a `sha256sum` on the project files?), but I think it would be best to append it with a `*` so that we know which commit it was based on.

In fact, instead of doing

```make
GIT_VERSION := $(addprefix 0x,$(shell git describe --abbrev=4 --always))
```
I would just do

```make
GIT_VERSION := $(shell git describe --tags --abbrev=8 --always --dirty=*)
```
This gives an output like: `v1.0-1-g8de68bca*`.

The `v1.0` is the tag, the `-1` tells us how many commits off of `v1.0` this is, and the `-g` tells us that it came from a git VCS. `--dirty=*` takes care of appending the asterisk if the repo is dirty. *Note that git does not consider having untracked files as being dirty*.

Also, I changed it to `--abbrev=8` because as the number of commits increase, generally the more hash characters are needed to identify the commit uniquely.

If this command is run in a repo with no tags, you will just get: `8de68bca*`

For what it's worth, I would vote the heartbeat since that's [where](https://github.com/byu-magicc/mavlink_c_library/blob/master/common/mavlink_msg_heartbeat.h#L12) MAVLink puts its version.

superjax:
I guess that makes sense.   I was trying to make it fit in a uint32 rather than a string, because the heartbeat doesn't have a string. But a string would give us a lot more freedom and allow us to put in tags.

We would just have to create a new message and send the version on that.   We could even use the debug messages to send it.

dpkoch:
The param read function is currently broken. Everything else seems to be working, I've been poking around with the debugger but haven't yet figured out what the issue is. It seems like in the `read_params()` function the call to `memory_read()` is returning false, but that function is hard-coded to return true. Still trying to figure out what's going on there.

Also noticed that the `sizeof(params_t)` call is returning a size of 2152, but adding up the contents of that struct manually the size is 2151. I think what's going on is that the compiler is padding an extra byte after the `version`, `size`, and `magic_be` fields (total of 7 bytes) so that the arrays start on the beginning of a 32-bit word. Before this change, those fields were a total of 4 bytes, so this padding wouldn't have been necessary. Don't know if that could be an issue, but it's also just speculation that this is happening and I haven't confirmed it.

@superjax Do you have any thoughts off the top of your head?

dpkoch:
Just kidding. Stray semicolon in 373b075... curse you @superjax! :) Fixed now.

dpkoch:
Closes #143

superjax:
Oops.  Sorry about that.  Thanks for catching it

superjax:
Tested on hardware.  Seems to be working properly.

Just FYI.  It seems that you have to do a rebuild in order to update the git hash.  I had a dirty commit, and then committed it, and no files changes.  Therefore the makefile didn't know to re-compile.  After recompiling though it updated the hash.

PR #155, opened 2017-05-08
[WIP] Error state

superjax:
I accidentally based this off the `documentation` branch.  Oops.  We need to do that merge first, or perhaps do a cherry-pick.

This still needs to be tested on hardware.

PR #159, opened 2017-05-10
Error state (from RC1.0)

dpkoch:
This pull commit supersedes #155; it just cherry picks the commit from that one so that it's based off of the `RC1.0` branch instead of `documentation`.

### DIFF
--- a/src/sensors.c
+++ b/src/sensors.c
@@ -105,7 +105,10 @@ void init_sensors(void)
 
 bool update_sensors()
 {
-  // Look for disabled sensors while disarmed (poll every 0.5 seconds)
+  // First, check for new IMU data
+  bool new_imu_data = update_imu();
+
+  // Now, look for disabled sensors while disarmed (poll every 0.5 seconds)
   // These sensors need power to respond, so they might not have been
   // detected on startup, but will be detected whenever power is applied
   // to the 5V rail.
@@ -140,6 +143,7 @@ bool update_sensors()
     }
   }
 
+  // Update whatever sensos are available
   if(baro_present())
   {
     baro_read(&_baro_altitude, &_baro_pressure, &_baro_temperature);
@@ -170,7 +174,7 @@ bool update_sensors()
   }
 
   // Return whether or not we got new IMU data
-  return update_imu();
+  return new_imu_data;
 }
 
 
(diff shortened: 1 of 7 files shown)


src/sensors.c
```diff
@@ -96,6 +96,8 @@ static bool update_imu(void);
 // function definitions
 void init_sensors(void)
 {
+  // clear the IMU read error
+  _error_state &= ~(ERROR_IMU_NOT_RESPONDING);
```
dpkoch: Right now it looks like the `ERROR_IMU_NOT_RESPONDING` flag doesn't get set anywhere?

superjax:
I think you're right.  I was actually thinking about this while I was laying in bed sick, and I was thinking that it should probably be a state machine, like you said.  I was thinking also that the status message should get published every time that an armed state, error state or any of the other "flag" members of the status message gets changed.

i agree, that is future work, but I think that's the answer.

I'll incorporate the prevention of arming if we have an error state, and make sure it's being published on the ROS network, but I think that's as far as I'll go with the error state for this release.

src/sensors.c
superjax: I made a mistake, line 254 was supposed to set the error.

Unfortunately, it gets reset in line 100 during the `imu_not_responding_error()` called in line 255.  That means it basically doesn't ever get called.   However, looking forward to our state machine setup, I hope that it gets published at least once so we know that something happened even if it gets fixed almost immediately.

superjax:
I tested this at my desk.  Seems to be working properly.  Requires an update to fcu_io from the `error_state` branch

dpkoch:
Cool, I like this, and I agree with that plan. I seem to be getting a `TIME_GOING_BACKWARDS` error every time, does that happen to you?

superjax:
I actually haven't gotten it once yet.

superjax:
Ah! just got one.

superjax:
This is weird...

image

Off by a single microsecond?  That is straight-up weird.

Well, unless it's really off by 999 microseconds, and the millisecond timer didn't increment.  I guess that would make sense.

superjax:
Okay, I made a change to breezy that puts highest priority on systick interrupts.  I2C are the next highest.  I ran this on my desk for about 30 minutes and saw 4 I2C errors in the first second, but then none after that.  Loop rate seemed stable and I didn't get any `TIME_GOING_BACKWARDS` errors.

dpkoch:
Awesome, nice work! Looks like this error message stuff has already paid off

dpkoch:
One last question. Right now if we do get a `TIME_GOING_BACKWARDS` error, that flag never gets unset and so we could never recover and arm without rebooting. Is that the behavior we want, or should that flag get cleared somewhere?

superjax:
Hmm.  Probably not.  I guess we could reboot to clear the error, but then that seems extreme.  What conditions would cause it to clear?  We could allow some error codes to arm, and others not to.

superjax:
Okay, we are now just clearing that error the next time that the estimator runs properly.  Ready to merge!

PR #168, opened 2017-06-09
Removed gyro calibration when arming

### LINKED ISSUE #166: Shifted Gyro Bias
The Gyro bias changed mid-flight  while flying ROS Plane. 
bad_gyro_bias

dpkoch:
Removed gyro calibration when arming. Tested at my desk and it arms like it's supposed to. Haven't actually verified that the gyro doesn't recalibrate, but it seems pretty obvious that it won't.

Closes #166

### DIFF
--- a/src/mode.c
+++ b/src/mode.c
@@ -1,22 +1,22 @@
-/* 
+/*
  * Copyright (c) 2017, James Jackson and Daniel Koch, BYU MAGICC Lab
- * 
+ *
  * All rights reserved.
- * 
+ *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


dpkoch:
I was able to reproduce the issue in #166 with the code right before this pull request, then demonstrate that this fixes it.

I calibrated the gyros with the naze stationary, then rotated it while arming (slowly enough that the "too much movement" flag didn't get thrown). With the previous code you can see that I produce an offset similar in magnitude to what we saw in #166, while with the new code in this pull request no offset occurs.

Before:
before_no_gyro_calib

After:
after_no_gyro_calib

PR #177, opened 2017-06-14
Removed outdated info from README file

dpkoch:
Eventually we'll probably want to strip out even more of the stuff in the README file and move into a "developer guide" section in the documentation, but for now I just removed a few things that were outdated.


superjax:
I agree.  Thanks for fixing that.

PR #180, opened 2017-06-16
simple lidar implementation

len0rd:
Added simple lidar support using the sen13680 driver. The measurements are not yet used anywhere else in the firmware, but they do publish across mavlink correctly.


len0rd:
@superjax @dpkoch On board.lidar_present(), I left the check for sen13680_read() < 257. Without it, if external power is supplied to the board after usb power, the lidar wont initialize and will just hold the same value forever. This is a pretty specific scenario, so if you thing the condition isn't warranted, I can remove it.

dpkoch:
Superseded by the C++ stuff.

PR #181, opened 2017-06-21
[WIP] C++ Port

superjax:
Here is the first hack at a C++ port to ROSflight.  This should be completely up to date with v0.1.0, but it needs to be flown on hardware, and the architecture needs to change somewhat per my discussion with @dpkoch  and @gellings. 

I figured that this is probably about 90% of the work already done.  It builds and at least runs on the naze32.

All that is left I believe:

- [x] Move to each module each having a reference to the rosflight parent instead of a bunch of pointers

- [x] Switch to a protected data struct for information passing between modules.

- [ ] Fixing mavlink_log messages so they work in the main code body

- [x] Setting up a testing framework

- [x] Making the state manager in charge of motor commands and rosflight_cmd messages

- [x] Renaming things to match the naming convention we came up with in our discussions


superjax:
Okay, @dpkoch we have one problem with the proposed method.

Right now, the PID controllers are super efficient because they directly access the memory to the command structs for output.  Is there a way we can keep that exposed in the new framework?

dpkoch:
Hey @superjax I think you forgot to commit the new `src/mavlink.cpp` file in that last commit. Could you add that?

superjax:
Okay, I found an issue that I don't know how to solve.

The way our code currently works, it's impossible to know if we are correctly matching the IMU timestamp with the data it's associated with.  That's why we are getting the TIME_GOING_BACKWARDS errors.  Sometimes, the time gets incremented before we can handle the data, and we associate old data with a new timestamp, which means that the next time we run the estimator, it tells us we have new IMU data, but it has the same timestamp.  Kind of frustrating.

I think the right way to do this is to do asynchronous IMU updates.  I think I know enough about how the NVIC works on the F1 that we can make this work now.

superjax:
Closes #183, Mostly finishes #184, #186 and #187

Issue #195, opened 2017-06-28
C++ Unit Tests


superjax:

superjax:
- [ ]  State Manager

- [ ]  Command Manager

- [ ]  RC override

- [ ]  Turbotrig/Vector

superjax:
Closed by #197

PR #197, opened 2017-07-01
Test

superjax:
Unit tests for math library, state machine, rc override and command manager


superjax:
Closes #195

test/command_manager_test.cpp
```diff
@@ -0,0 +1,610 @@
...
+  //=================================================
+
+  // Only override attitude
+  rf.params_.set_param_int(PARAM_RC_ATTITUDE_OVERRIDE_CHANNEL, 4);
+  rf.params_.set_param_int(PARAM_RC_THROTTLE_OVERRIDE_CHANNEL, 4);
+
+  // switch is off, but roll channel is deviated
+  rc_values[4] = 1100;
+  rc_values[0] = 1800;
+  rc_values[2] = 1000;
```
dpkoch: What is `rc_values[2]` doing?

test/command_manager_test.cpp
```diff
@@ -0,0 +1,610 @@
...
+  rf.command_manager_.set_new_offboard_command(offboard_command);
+  step_firmware(rf, board, 20000);
+  output = rf.command_manager_.combined_control();
+  EXPECT_PRETTYCLOSE(output.F.value, 0.2);
+
+
+
+  // Okay, remove the Min throttle setting
+  rf.params_.set_param_int(PARAM_RC_OVERRIDE_TAKE_MIN_THROTTLE, false);
+
```
dpkoch: Let's add a test here to make sure that min throttle is actually off now?

test/command_manager_test.cpp
```diff
@@ -0,0 +1,610 @@
...
+  float max_roll = rf.params_.get_param_float(PARAM_RC_MAX_ROLL);
+  float max_pitch = rf.params_.get_param_float(PARAM_RC_MAX_PITCH);
+  float max_yawrate = rf.params_.get_param_float(PARAM_RC_MAX_YAWRATE);
+
+  // Let's clear all errors in the state_manager
+  rf.state_manager_.clear_error(rf.state_manager_.state().error_codes);
+
+  //=================================================
+  // RC Arming Test
+  //=================================================
```
dpkoch: I wonder if these arming tests should get moved into the state manager tests?

test/command_manager_test.cpp
```diff
@@ -0,0 +1,610 @@
...
+  EXPECT_EQ(rf.state_manager_.state().failsafe, false);
+
+  // flip the arm switch off
+  rc_values[4] = 1100;
+  board.set_rc(rc_values);
+  step_firmware(rf, board, 20000);
+  // we should be disarmed
+  EXPECT_EQ(rf.state_manager_.state().armed, false);
+  EXPECT_EQ(rf.state_manager_.state().error, false);
+  EXPECT_EQ(rf.state_manager_.state().failsafe, false);
```
dpkoch: Let's also add the following tests:
  - sticks take the appropriate amount of time to arm
  - sticks don't arm when they shouldn't
  - switch doesn't arm when it shouldn't (maybe just set all of the switch channels high to cover them all and make sure it doesn't arm)

test/command_manager_test.cpp
superjax: We could do that.  I figured that they were extensive enough that they justified their own file.  They used to be a bunch of separate executables, but I combined them to make it easier to run the test.

superjax: I'm mostly concerned with how tong the ifles get

test/command_manager_test.cpp
superjax: cool

test/command_manager_test.cpp
```diff
@@ -0,0 +1,610 @@
...
+  while (board.clock_micros() < 1000000)
+  {
+    step_firmware(rf, board, 20000);
+  }
+
+  offboard_command.stamp_ms = board.clock_millis();
+  rf.command_manager_.set_new_offboard_command(offboard_command);
+  step_firmware(rf, board, 20000);
+
+  // We don't have an override channel mapped, so this should be RC
```
superjax: yup

test/command_manager_test.cpp
superjax: moving the throttle to neutral position

dpkoch:
Our Travis builds should run a little faster now; turns out it by default assumes you're using ruby and downloads and installs a bunch of ruby stuff. By specifying C++ as the language it skips all of that.

superjax:
Cool!  Thanks

Issue #198, opened 2017-07-06
rc_override doesn't activate with x/y deviation


len0rd:
While using the new status message, I noticed that when in autopilot mode, rc_override would only switch to true if I deviated yaw, but not if I deviated roll/pitch. This would be hard to catch without the fixed status message, since the old message would not accurately display rc_override state.

superjax:
This is addressed by #199

len0rd:
pretty sure this is fixed in C++ implementation

PR #199, opened 2017-07-06
Fix corresponding to issue #198

len0rd:
Added local rc_ovrd variable to do_roll_pitch_yaw_muxing to mirror do_throttle muxing. Also changed these local variable names for clarity, so they cant be confused with the static rc_override in the class. The mux loop in the mux_inputs function no uses the static rc_override data member.


src/mux.c
```diff
@@ -175,36 +175,37 @@ static bool stick_deviated(mux_channel_t channel)
 
 static bool do_roll_pitch_yaw_muxing(mux_channel_t channel)
 {
+  bool rc_ovrd;
   //Check if the override switch exists and is triggered, or if the sticks have deviated enough to trigger an override
   if ((rc_switch_mapped(RC_SWITCH_ATT_OVERRIDE) && rc_switch(RC_SWITCH_ATT_OVERRIDE)) || stick_deviated(channel))
   {
-    rc_override = true;
```
superjax: Why did you rename this variable?

len0rd: Just for clarity, since the original issue was that function was writing to the static rc_override variable when it should have been writing to the local rc_override variable that it returns, as is done in the throttle muxing function. I just found it confusing having the local and static variables with the same name, so I decided to change it.

superjax:
I think that this has been fixed in the latest rc1.0

PR #208, opened 2017-07-17
Log support

superjax:
slight modification to extra sensor support, added log message support back to main body of code.


src/mavlink.cpp
```diff
@@ -416,37 +427,37 @@ void Mavlink::send_attitude(void)
...
+                             gyro.y,
+                             gyro.z,
+                             RF_.sensors_.data().imu_temperature);
+  send_message(msg);
+  //  }
+  //  else
+  //  {
+  // Otherwise, wait and signal that we still need to send IMU
+  //    mavlink_streams[STREAM_ID_IMU].next_time_us -= mavlink_streams[STREAM_ID_IMU].period_us;
+  //  }
```
dpkoch: Just double checking, is this stuff supposed to be commented out?

superjax: Yeah, that is on purpose (for now).   That little routine is to make sure that we don't send the same IMU measurement more than once.

src/sensors.cpp
```diff
@@ -59,80 +59,115 @@ void Sensors::init()
...
+      float mag[3];
+      rf_.board_.mag_read(mag);
+      data_.mag.x = mag[0];
+      data_.mag.y = mag[1];
+      data_.mag.z = mag[2];
+      correct_mag();
+    }
+    break;
   }
+  next_sensor_to_update_ = (next_sensor_to_update_ + 1) % 4;
```
superjax: Yeah.  I do too.  It just seemed like a lot of overhead at the time.  Perhaps it would make it cleaner

superjax:
Yeah, the sensor streaming still works.

PR #212, opened 2017-07-18
Standard motor numbering

len0rd:
Changed mixer numbering to a normal clockwise pattern for quadx, quad+ and x8 types. Updated the hardware-setup doc with new images, also fixed mixer-value table for hardware-setup as it was incorrect. QuadX and quad+ tested on thor and work. X8 is UNTESTED.


len0rd:
For this to work correctly on thor though we'd have to switch the esc plug

PR #221, opened 2017-07-24
Fix 202

superjax:
Here are a few fixes from flight testing on Friday, and fixing unit test issues.


superjax:
I don't know why it is failing on Travis.  It works on my machine.  I'll work on that later.

superjax:
This should be ready to go!

src/estimator.cpp
```diff
@@ -257,7 +267,8 @@ void Estimator::run()
 
   // If it has been more than 0.5 seconds since the acc update ran and we are supposed to be getting them
   // then trigger an unhealthy estimator error
-  if (RF_.params_.get_param_int(PARAM_FILTER_USE_ACC) && now_us > 500000 + last_acc_update_us_)
+  if (RF_.params_.get_param_int(PARAM_FILTER_USE_ACC) && now_us > 500000 + last_acc_update_us_
+      && RF_.params_.get_param_int(PARAM_FIXED_WING))
```
dpkoch: Shouldn't this be a "not" on the fixed wing?

src/param.cpp
```diff
@@ -164,17 +164,17 @@ void Params::set_defaults(void)
...
-  init_param_float(PARAM_FILTER_KI, "FILTER_KI", 0.1f); // estimator integral gain - See estimator documentation | 0 | 1.0
+  init_param_float(PARAM_FILTER_KP, "FILTER_KP", 2.0f); // estimator proportional gain - See estimator documentation | 0 | 10.0
+  init_param_float(PARAM_FILTER_KI, "FILTER_KI", 0.2f); // estimator integral gain - See estimator documentation | 0 | 1.0
 
-  init_param_int(PARAM_FILTER_USE_QUAD_INT, "FILTER_QUAD_INT", 0); // Perform a quadratic averaging of LPF gyro data prior to integration (adds ~20 us to estimation loop on F1 processors) | 0 | 1
-  init_param_int(PARAM_FILTER_USE_MAT_EXP, "FILTER_MAT_EXP", 0); // 1 - Use matrix exponential to improve gyro integration (adds ~90 us to estimation loop in F1 processors) 0 - use euler integration | 0 | 1
+  init_param_int(PARAM_FILTER_USE_QUAD_INT, "FILTER_QUAD_INT", 1); // Perform a quadratic averaging of LPF gyro data prior to integration (adds ~20 us to estimation loop on F1 processors) | 0 | 1
+  init_param_int(PARAM_FILTER_USE_MAT_EXP, "FILTER_MAT_EXP", 1); // 1 - Use matrix exponential to improve gyro integration (adds ~90 us to estimation loop in F1 processors) 0 - use euler integration | 0 | 1
```
dpkoch: Just double checking this isn't left over from debugging, you meant to turn these on by default?

include/controller.h
```diff
@@ -89,9 +89,10 @@ class Controller
     float tau_;
   };
 
-  vector_t run_pid_loops(float dt, const Estimator::State& state, const control_t& command, bool update_integrators);
-
   ROSflight& RF_;
+
+  vector_t run_pid_loops(int32_t dt, const Estimator::State& state, const control_t& command, bool update_integrators);
```
dpkoch: Are we going to allow negative dts? If not maybe we should change dt to a uint32_t to enforce that

src/controller.cpp
```diff
@@ -96,22 +96,26 @@ void Controller::init()
 void Controller::run()
 {
   // Time calculation
-  if (prev_time_ < 0.0000001)
+  if (prev_time_us_ < 1)
```
dpkoch: Can this just be a `if (prev_time_us_ == 0)`?

src/param.cpp
superjax: I did.  We have enough processor time that there isn't any real reason to not use them (unit tests also all work for them too, so we know they work)

src/estimator.cpp
superjax: Haha!  Yep.  Thanks.

PR #228, opened 2017-08-02
Sensor outlier rejection

drinkdhmo:
Added outlier rejection for baro, diff pressure, and sonar measurements using a sliding window filter. Also added online variance estimation during calibration so that the calibration will fail when the sensor measurements aren't stable enough.

### DIFF
--- a/include/sensors.h
+++ b/include/sensors.h
@@ -54,12 +54,15 @@ class Sensors
     float diff_pressure_velocity = 0;
     float diff_pressure = 0;
     float diff_pressure_temp = 0;
+    bool diff_pressure_valid = false;
 
     float baro_altitude = 0;
     float baro_pressure = 0;
     float baro_temperature = 0;
+    bool baro_valid = false;
 
     float sonar_range = 0;
+    bool sonar_range_valid = false;
 
     vector_t mag = {0, 0, 0};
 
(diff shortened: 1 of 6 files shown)


include/sensors.h
```diff
@@ -94,6 +98,28 @@ class Sensors
...
+  static const float BARO_MAX_CHANGE_RATE;
+  static const float BARO_SAMPLE_RATE;
+  static const float DIFF_MAX_CHANGE_RATE;
+  static const float DIFF_SAMPLE_RATE;
+  static const float SONAR_MAX_CHANGE_RATE;
+  static const float SONAR_SAMPLE_RATE;
+  static const int SENSOR_CAL_DELAY_CYCLES;
+  static const int SENSOR_CAL_CYCLES;
+  static const float BARO_MAX_CALIBRATION_VARIANCE;
+  static const float DIFF_PRESSURE_MAX_CALIBRATION_VARIANCE;
```
superjax: Hey @drinkdhmo,  So, I'm wondering why have these `const`s.  Is it just for clarity?  Should they be parameters?

drinkdhmo: I did it for clarity. We could make at least the max variances a parameter for someone trying to get an especially bad sensor to work, but I don't think its necessary.

drinkdhmo:
This branch was successfully flight tested yesterday.

PR #245, opened 2017-08-11
compiling in strict mode

### LINKED ISSUE #227: "Strict Compilation"
Per request by @sytelus,  it would be nice if we could adhere to strict compilation.

Here is a list of @sytelus's compiler flags in CMake syntax

`set(CMAKE_CXX_FLAGS "-std=c++11 -Wall -Wextra  -Wstrict-aliasing -pedantic -fmax-errors=5 -Werror -Wunreachable-code -Wcast-align -Wcast-qual -Wctor-dtor-privacy -Wdisabled-optimization -Wformat=2 -Winit-self -Wlogical-op -Wmissing-include-dirs -Wnoexcept -Wold-style-cast -Woverloaded-virtual -Wredundant-decls -Wshadow -Wsign-promo -Wstrict-null-sentinel -Wstrict-overflow=5 -Wswitch-default -Wundef -Wno-unused -Wno-variadic-macros -Wno-parentheses -fdiagnostics-show-option ${CMAKE_CXX_FLAGS}")`

I tried this, and for the most part, it's mad at us for some of the legacy type castings we have lying about.  It might be a good idea to update those.

[Link to @sytelus blog](http://shitalshah.com/p/how-to-enable-and-use-gcc-strict-mode-compilation/)

superjax:
Closes #227

### DIFF
--- a/include/param.h
+++ b/include/param.h
@@ -294,14 +294,17 @@ public:
    * @param id The ID of the parameter
    * @return The value of the parameter
    */
-  inline const int get_param_int(uint16_t id) const { return params.values[id]; }
+  inline int get_param_int(uint16_t id) const { return params.values[id]; }
 
   /**
    * @brief Get the value of a floating point parameter by id
    * @param id The ID of the parameter
    * @return The value of the parameter
    */
-  inline const float get_param_float(uint16_t id) const { return *(float *) &params.values[id]; }
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstrict-aliasing"
+  inline float get_param_float(uint16_t id) const { return reinterpret_cast<const float &>(params.values[id]); }
+#pragma GCC diagnostic pop
 
   /**
    * @brief Get the name of a parameter
@@ -318,7 +321,7 @@ public:
    * PARAM_TYPE_INT32, PARAM_TYPE_FLOAT, or PARAM_TYPE_INVALID
    * See line 165
    */
-  inline const param_type_t get_param_type(uint16_t id) const { return params.types[id]; }
+  inline param_type_t get_param_type(uint16_t id) const { return params.types[id]; }
 
   /**
    * @brief Sets the value of a parameter by ID and calls the parameter change callback
--- a/include/param.h
+++ b/include/param.h
@@ -221,13 +221,19 @@ public:
   static constexpr uint8_t PARAMS_NAME_LENGTH = 16;
 
 private:
+  union param_value_t
+  {
+    float fvalue;
+    int32_t ivalue;
+  };
+
   typedef struct
   {
     uint32_t version;
     uint16_t size;
     uint8_t magic_be;                       // magic number, should be 0xBE
 
-    int32_t values[PARAMS_COUNT];
+    param_value_t values[PARAMS_COUNT];
     char names[PARAMS_COUNT][PARAMS_NAME_LENGTH];
     param_type_t types[PARAMS_COUNT];
 
@@ -294,17 +300,14 @@ public:
    * @param id The ID of the parameter
    * @return The value of the parameter
    */
-  inline int get_param_int(uint16_t id) const { return params.values[id]; }
+  inline int get_param_int(uint16_t id) const { return params.values[id].ivalue; }
 
   /**
    * @brief Get the value of a floating point parameter by id
    * @param id The ID of the parameter
    * @return The value of the parameter
    */
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wstrict-aliasing"
-  inline float get_param_float(uint16_t id) const { return reinterpret_cast<const float &>(params.values[id]); }
-#pragma GCC diagnostic pop
+  inline float get_param_float(uint16_t id) const { return params.values[id].fvalue; }
 
   /**
    * @brief Get the name of a parameter
--- a/lib/turbotrig/turbotrig.cpp
+++ b/lib/turbotrig/turbotrig.cpp
@@ -233,7 +233,7 @@ float turboatan(float x)
     return M_PI/2.0 - turboatan(1.0/x);
   }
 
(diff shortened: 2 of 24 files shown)


superjax:
Done!

include/param.h
```diff
@@ -294,14 +294,17 @@ class Params
...
 
   /**
    * @brief Get the value of a floating point parameter by id
    * @param id The ID of the parameter
    * @return The value of the parameter
    */
-  inline const float get_param_float(uint16_t id) const { return *(float *) &params.values[id]; }
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstrict-aliasing"
+  inline float get_param_float(uint16_t id) const { return reinterpret_cast<const float &>(params.values[id]); }
```
dpkoch: I'm going to see if I can fix this by switching the param stuff to using a `union` instead of casting... I'd like to get rid of as many of these `GCC diagnostic ignored` sections as possible. Ideally, I'd like to get rid of all of them in our flight stack and only use it for mavlink and breezy.

lib/turbotrig/turbotrig.cpp
```diff
@@ -233,16 +233,16 @@ float turboatan(float x)
     return M_PI/2.0 - turboatan(1.0/x);
   }
 
-  float t = (x - atan_min_x)/(float)(atan_max_x - atan_min_x) * (float)atan_num_entries;
-  int16_t index = (int)t;
-  float dx = t - index;
+  float t = (x - atan_min_x)/static_cast<float>(atan_max_x - atan_min_x) * static_cast<float>(atan_num_entries);
```
dpkoch: `atan_max_x` and `atan_min_x` are both already floats, we shouldn't need this cast

lib/turbotrig/turbotrig.cpp
```diff
@@ -293,26 +293,26 @@ float turboasin(float x)
     return -1.0*turboasin(-1.0*x);
   }
 
-  float t = (x - asin_min_x)/(float)(asin_max_x - asin_min_x) * (float)asin_num_entries;
-  int16_t index = (int)t;
-  float dx = t - index;
+  float t = (x - asin_min_x)/static_cast<float>(asin_max_x - asin_min_x) * static_cast<float>(asin_num_entries);
```
dpkoch: `asin_max_x` and `asin_min_x` are both already floats, don't need this cast

lib/turbotrig/turbovec.cpp
```diff
@@ -172,6 +172,10 @@ void euler_from_quat(quaternion_t q, float *phi, float *theta, float *psi)
 }
 
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wold-style-cast"
+#pragma GCC diagnostic ignored "-Wcast-qual"
```
dpkoch: If there's any way we could get rid of these, that'd be awesome... I'll take a look at it

src/estimator.cpp
```diff
@@ -139,7 +138,7 @@ void Estimator::run()
   state_.timestamp_us = now_us;
 
   // Crank up the gains for the first few seconds for quick convergence
-  if (now_us < (uint64_t)RF_.params_.get_param_int(PARAM_INIT_TIME)*1000)
+  if (now_us < static_cast<uint64_t>(RF_.params_.get_param_int(PARAM_INIT_TIME)*1000))
```
dpkoch: Should the `*1000` be outside of the cast to `uint64_t`? Just wondering if there's possibility for overflow here if we do the multiplication on a `int32_t`

src/mavlink.cpp
```diff
@@ -157,11 +157,16 @@ void Mavlink::handle_msg_param_set(const mavlink_message_t *const msg)
         switch (candidate_type)
         {
         case PARAM_TYPE_INT32:
-          RF_.params_.set_param_int(id, *(int32_t *) &set.param_value);
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstrict-aliasing"
+          RF_.params_.set_param_int(id, reinterpret_cast<const int32_t &>(set.param_value));
```
dpkoch: I'll see if I can fix this using a `union` or something

src/nanoprintf.cpp
```diff
@@ -31,6 +31,12 @@
 
 #include "nanoprintf.h"
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wold-style-cast"
+#pragma GCC diagnostic ignored "-Wcast-qual"
+#pragma GCC diagnostic ignored "-Wstrict-overflow"
```
dpkoch: I'm guessing there's no way we're going to get this file cleaned up... oh well.

src/rc.cpp
```diff
@@ -267,21 +265,21 @@ bool RC::run()
...
     uint16_t pwm = RF_.board_.pwm_read(sticks[channel].channel);
     if (sticks[channel].one_sided) //generally only F is one_sided
     {
-      stick_values[channel] = ((float)(pwm - 1000)) / 1000.0f;
+      stick_values[channel] = (static_cast<float>(pwm - 1000)) / 1000.0f;
     }
     else
     {
-      stick_values[channel] = (float)(2*(pwm - 1500)) / (1000.0);
+      stick_values[channel] = static_cast<float>(2*(pwm - 1500)) / (1000.0);
```
dpkoch: Missing a float qualifier here on the `1000.0`

src/nanoprintf.cpp
superjax: Haha, yeah, unless we re-code the whole thing.

src/estimator.cpp
superjax: I think it makes sense, from a clarity point of view, but it's not likely that the user will specify an initialization time of more than 20 minutes (the overflow of int32)

superjax:
Alright, I got rid of all compiler warnings in the main flight stack, except for the inclusion of mavlink, which should eventually get hidden with #248

PR #248, opened 2017-08-14
Abstract comm link layer

### LINKED ISSUE #188: C++ CommLink Abstraction
This is actually pretty tricky.  First, let's define the API that we need for a CommLink abstraction layer, and then implement it.

dpkoch:
Closes #188


superjax:
I'm not able to test this, but I feel like it should be flown before merging.

dpkoch:
@superjax Agreed. I've just bench-tested it so far, but it seemed to work fine. I still have some clean-up to do before this is ready to merge.

dpkoch:
Okay @superjax, I think this is good enough to merge. There are still a few minor things I'd like to tweak (like more strict use of enum values when appropriate), but it works and I'd rather not get too far behind master again. In order to get rid of the interdependency, I had to pass in the baud rate as a parameter to the `CommLink` base class constructor. This kind of assumes that all derived `CommLink` implementations will be serial protocols which obviously isn't true, but it's the best idea I have at the moment. For now if you're implementing a version that does direct function calls then you can just ignore that parameter. Let me know what you think.

This has been bench tested and seems to be working, but we should probably fly before mergin.

superjax:
Wow!  I'll go through this when I have time.  This is awesome!

superjax:
This is awesome!

Do we want to move the `mavlink.cpp` and `mavlink.h` into the `lib/mavlink` folder?

dpkoch:
@superjax It probably makes sense to move them somewhere out of the main `include` and `src` folders, but I'm trying to decide if `lib/mavlink` is the right place. It could work, but it seems a little strange to mix some of our code with a bunch of external code and lump it all together as the `mavlink` library (since the external mavlink library is already a full-blown, self-contained library).

Another option would be to move them into a new `comms` folder, like what we with the `boards` folder. Since both boards and comm layers are things that can be extracted but the implementations aren't really part of the flight stack, it might make sense to give them a parallel directory structure. That also kind of tucks the mavlink stuff out of the way nicely if you're not using it. On the other hand, boards have their own makefiles and stuff so it makes some sense for them to get preferential treatment and maybe slightly less so for comm layers.

Here are the options:

Option 1 (`lib` folder):
```
|- include/
|- src/
|- boards/
|   |- naze/
|       |- meh.cpp
|       |- breezystm32/
|- lib/
    |- mavlink/
    |   |- mavlink.h
    |   |- mavlink.cpp
    |   |- v1.0/
    |       |- all the stuff
    |- turbomath/
```
Option 2 (`comms` folder):
```
|- include/
|- src/
|- boards/
|   |- naze/
|       |- meh.cpp
|       |- breezystm32/
|- comms/
|   |- mavlink/
|       |- mavlink.h
|       |- mavlink.cpp
|       |- v1.0/
|           |- all the stuff
|- lib/
    |- turbomath/
```

What are your thoughts? I'm guessing you never really wanted to think about it this much in the first place...

dpkoch:
I think I like the `comms` folder idea. If you're okay with that @superjax I'll get it all set up

superjax:
I like the `comms` folder idea too.  I like how it is parallel to the `boards` abstraction.

dpkoch:
Alright @superjax, I've got the remaining MAVLink dependencies removed and the mavlink comm layer moved into the `comms` directory like we decided.

The biggest things that I refactored were logging, commands (like calibrate IMU), and offboard control. I've bench tested the logging and commands and they seem to work, but I haven't tested offboard control yet. We should probably flight test that before merging. It'd probably also be good to write some unit tests for the comm layer.

superjax:
This was tested on Leo

PR #251, opened 2017-08-30
added from_RPY, reduced size of lookup tables with clever scaling, an…

superjax:
…d added tests

Pull Request Guidelines

all pull requests should adhere to the [style guide](http://docs.rosflight.org/en/developer_docs/developer-guide/style_guide/)

and should pass the unit tests

``` bash
./run_tests.sh
```
Any pull request not adhering to these guidelines will be rejected.


lib/turbomath/turbomath.h
```diff
@@ -67,6 +67,10 @@ class Vector
   Vector& operator-= (const Vector& v);
 };
 
+inline Vector operator* (float s, Vector v) { return v * s; }
+inline Vector operator/ (float s, Vector v) { return v / s; }
+
```
dpkoch: Let's make these `const Vector&`

PR #255, opened 2017-09-26
Fixed min throttle default, added tests for default parameters

dpkoch:
- Change the default for taking minimum of RC and offboard throttle to true
 - Add unit tests for default values of important parameters


dpkoch:
I also propose that we cherry-pick this commit onto the v1.0 branch and release it as v1.0.1. Do you agree @superjax?

superjax:
I agree.

PR #260, opened 2017-09-29
fixed #258

### LINKED ISSUE #258: Problems With rev5 Naze
The IMU not responding error is never cleared and it doesn't look like it properly switches to rev5 after realizing that it hasn't gotten IMU in a while.

superjax:
Fixed bug


superjax:
I think this should be pulled into stable as quickly as possible.

PR #267, opened 2017-11-10
update hardware-setup doc to clarify channel to servo and motor mapping. Thus, hopefully users won't connect a servo to the motor pins, for example.

uavteambyu:
Pull Request Guidelines

all pull requests should adhere to the [style guide](http://docs.rosflight.org/en/developer_docs/developer-guide/style_guide/)

and should pass the unit tests

``` bash
./run_tests.sh
```
Any pull request not adhering to these guidelines will be rejected.

### DIFF
--- a/docs/user-guide/hardware-setup.md
+++ b/docs/user-guide/hardware-setup.md
@@ -114,3 +114,20 @@ The _ESC calibration_ mixer outputs the throttle command equally to each motor, 
 ![Mixer_1](images/mixers_1.png)
 
 ![Mixer_2](images/mixers_2.png)
+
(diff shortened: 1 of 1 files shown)


uavteambyu:
Has anybody looked at this yet?

superjax:
Hey yeah.  I've looked at it.  Unfortunately, I don't know that it's actually helpful for all platforms.  The port difference between servos and escs are dependant on which mixer is used.  For example, a tricopter uses a a different port for its servo than a fixed wing.  That limits the applicability of the note to a single mixer, which is already pretty clear from the mixer diagram.

PR #283, opened 2018-02-21
[WIP] fixed controller bug

superjax:
This needs to be tested on multirotor hardware


superjax:
Fixed D term - tested on hardware

PR #284, opened 2018-02-21
Fix 148

superjax:
This has been tested and needs to be merged


include/sensors.h
```diff
@@ -80,7 +80,9 @@ class Sensors
 
   // function declarations
   void init();
+  void init_imu();
```
dpkoch: Make this private?

PR #286, opened 2018-02-23
[WIP] Aux Command

superjax:
This is a branch that @cmcquinn was working on.  It looks like it's ready to go, perhaps we should test it an merge it in?


dpkoch:
@austinhurst and @jhb1993 This is the aux servo command stuff that should work for your servo drop

superjax:
@cmcquinn, it would be great to get this merged in.  Do you think you have some time to help us get this synced back up with `master`?

cmcquinn:
Yes. I will get working on it.

superjax:
Hey @cmcquinn, I'm cleaning up the repo a bit, and I'm wondering what the status of this work is.

willisj2:
* Tested aux_command by publishing on the /aux_command and viewing the values passed through to /output_raw. 
* Fixed some issues with matching array sizes - all output_raw related array references are now for length 14.
* Modified mixer to include a PASSTHROUGH option - is the value chosen (11) acceptable or since this is more of a "special" option should it be grouped near ESC_CALIBRATION? The PWM update rate is set to 50 since this will likely be used with servos as well as ESCs.
* The mixer currently uses 8 outputs for mixing, the remaining 6 can be set only with the aux_command. Is this the way you would like it, or should the mixers be modified to mix all 14 outputs?

willisj2:
The aux command is working properly. This branch has been merged into gps because of dependency issues.

PR #294, opened 2018-05-14
Make `master` stable

superjax:
There are a lot of features that aren't in `stable` yet Primarily the rev 6 naze, and a few i2c fixes.


superjax:
Right, sorry I forgot how that all works... Let's release!  Do we need to do anything to test it?  (I really wish we had a fixedwing platform that we could test with)

PR #314, opened 2018-10-24
Serial Connection to F4

superjax:
This adds support to exposing the firmware over UART3 (the Flexi-port) on the F4.  This is useful for remote controlling the autopilot over a 3DR (SiK) radio.

Turns out, Xbee has really low bandwidth and is not suitable for real-time control with IMU streaming (I got at most 30 hz down)


.gitmodules
```diff
@@ -1,9 +1,15 @@
-[submodule "comms/mavlink/v1.0"]
-	path = comms/mavlink/v1.0
-	url = https://github.com/rosflight/mavlink_c_library.git
 [submodule "boards/breezy/breezystm32"]
 	path = boards/breezy/breezystm32
 	url = https://github.com/rosflight/BreezySTM32.git
 [submodule "boards/airbourne/airbourne"]
 	path = boards/airbourne/airbourne
 	url = https://github.com/rosflight/airbourne_f4.git
+[submodule "comms/mavlink/--force"]
```
dpkoch: Needs cleanup

boards/airbourne/airbourne_board.cpp
```diff
@@ -73,32 +75,49 @@ void AirbourneBoard::clock_delay(uint32_t milliseconds)
 }
 
 // serial
-void AirbourneBoard::serial_init(uint32_t baud_rate)
+void AirbourneBoard::serial_init(uint32_t baud_rate, uint32_t dev)
 {
-  (void)baud_rate;
   vcp_.init();
+  if (dev == 3)
```
dpkoch: Should we use an enum for the serial device IDs?

boards/airbourne/airbourne_board.cpp
```diff
@@ -73,32 +75,49 @@ void AirbourneBoard::clock_delay(uint32_t milliseconds)
...
 void AirbourneBoard::serial_write(const uint8_t *src, size_t len)
 {
-  vcp_.write(src, len);
+  current_serial_->write(src, len);
 }
 
 uint16_t AirbourneBoard::serial_bytes_available()
 {
-  return vcp_.rx_bytes_waiting();
+  if (vcp_.connected())
```
dpkoch: The logic seems to be slightly obfuscated in this implementation. If I'm correct, the big-picture logic is:

  * Use VCP if connected
  * If VCP is not connected _and_ another serial device has been set as the desired device, then use that device

I think we could streamline the implementation a bit to make this more clear?

boards/airbourne/airbourne_board.h
```diff
@@ -115,11 +121,13 @@ class AirbourneBoard : public Board
...
 
   // serial
-  void serial_init(uint32_t baud_rate) override;
+  void serial_init(uint32_t baud_rate, uint32_t dev) override;
   void serial_write(const uint8_t *src, size_t len) override;
   uint16_t serial_bytes_available() override;
   uint8_t serial_read() override;
   void serial_flush() override;
+  Serial** get_serial_interfaces(void);//Returns an array of pointers to available serial interfaces
+  uint8_t get_serial_count(void);//Returns the number of available serial interfaces. Same as length of array above.
```
dpkoch: Are these functions used/do they have an implementation?

boards/airbourne/airbourne_board.cpp
superjax: Yes, that is correct.  That would be better

boards/airbourne/airbourne_board.h
superjax: They are legacy and need to be removed

dpkoch:
I just did a little cleanup, deleted unused functions and now refer to serial devices by enum/switch statements (to get rid of magic numbers and make it expandable for other serial ports like the SBUS one). Still needs to be tested to make sure I didn't break anything.

Didn't change much and maybe wasn't worth being annoying about, but hopefully it's a little more clear what's going on?

The only other change I think we need before merging is adding documentation.

PR #317, opened 2018-10-26
Require low throttle and RC throttle override active to arm

### LINKED ISSUE #315: Switch arming works with non zero throttle
Hi all,

I have noticed a somewhat dangerous thing when arming using a switch; it allows arming even if the throttle is not close to zero.
If this the intended functionality? I would probably use the same threshold for throttle as the stick arming users to not have an accidental full throttle situation. :)

What are your thoughts on this?

BR Emil

dpkoch:
Closes #315

This hasn't been tested yet on hardware, will comment once that's been done

### DIFF
--- a/include/rc.h
+++ b/include/rc.h
@@ -98,8 +98,8 @@ private:
   rc_stick_config_t sticks[STICKS_COUNT];
   rc_switch_config_t switches[SWITCHES_COUNT];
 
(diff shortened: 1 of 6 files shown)


dpkoch:
Umm... this is weird. I don't know why one of the tests is passing and the other isn't. They pass on my computer.

PR #334, opened 2019-01-23
Issue 327 (Hardfault handling)

BillThePlatypus:
Add support for better hardfault handling on F4, including some backup memory.

### DIFF
--- a/include/board.h
+++ b/include/board.h
@@ -37,6 +37,8 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+#include "state_manager.h"
+
 namespace rosflight_firmware
 {
 struct debug_info_t{
@@ -53,10 +55,12 @@ struct backup_data_t{
     uint32_t error_code;
     debug_info_t debug_info;
     uint32_t reset_count;
-    uint32_t arm_status;
+    uint32_t arm_status; //This must equals ARM_MAGIC, or else the state manager will not rearm on reboot
     //TODO add state manager info
+    StateManager::State state;
     uint32_t checksum; //With the current implementation of the checksum, this must go last
 };
+const uint32_t ARM_MAGIC = 0xfa11bad;
 
 class Board
 {
--- a/include/board.h
+++ b/include/board.h
@@ -137,8 +137,8 @@ public:
   virtual void led1_toggle() = 0;
 
 // Backup memory
-  virtual bool has_error_data() = 0;
-  virtual backup_data_t get_error_data() = 0;
+  virtual bool has_backup_data() = 0;
+  virtual backup_data_t get_backup_data() = 0;
 
 };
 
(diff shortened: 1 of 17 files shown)


superjax:
Is this WIP or is this complete?

(If it's complete, then ask me or @dpkoch  to review it), if not, then append [WIP] to the beginning of the Issue name.

src/comm_manager.cpp
```diff
@@ -281,6 +282,19 @@ void CommManager::attitude_correction_callback(const turbomath::Quaternion &q)
 {
   RF_.estimator_.set_attitude_correction(q);
 }
+void CommManager::heartbeat_callback(void)
+{
+  static bool error_data_sent = false;
```
dpkoch: If I understand correctly, the idea behind this logic is that every time the flight controller starts up, when it receives its first heartbeat message from `rosflight_io` it will send backup data if it has it. This is the behavior we want immediately following a hard fault, but it seems like once something ever gets written to backup SRAM, this logic will always send that crash report after reboot, even if the system has been legitimately turned off and back on again, even multiple times. So do we need to delete or invalidate the contents of the backup SRAM after we send it to avoid that behavior?

superjax:
That logic occurs underneath the airborne layerhttps://github.com/rosflight/airbourne_f4/blob/7498992f7a1b458f59dd83ce277707895ae495ac/src/backup_sram.cpp#L34

dpkoch:
Okay. I think it'd be good if the board interface reflected that's what's happening. It'd probably be cleanest if we added a `clear_backup_data()` function and called it explicitly from inside the firmware. Having the logic split between the firmware and the board layer is a bit confusing and is likely to lead to errors if someone implements a new board layer later.

include/board.h
```diff
@@ -37,8 +37,31 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+#include "state_manager.h"
+
 namespace rosflight_firmware
 {
+struct debug_info_t{
```
dpkoch: Also, thinking into the future, this debug info struct is very specific to F4 boards; it kind of breaks the hardware abstraction and wouldn't apply if someone creates a new board support layer. Is there a way we can make it more generic? Maybe it's just a fixed-size array of `uint32_t` at the board layer, and then it needs to be reinterpreted based on what board it's actually running on? That keeps the abstraction cleaner but at the expense of it being a little more annoying for developers. This really shouldn't be used much though, and will only be used by people who have the background to understand how to interpret it. Thoughts?

BillThePlatypus: Most of this information is currently unused, as it is not very useful for debugging.  I did try to write it so that the backup_data_t was easy to modify, so any future additions/subtractions should be pretty seamless.

superjax: Maybe I'm missing something, but I don't see any F4-specific code here - (Every processor has a program counter, link register, etc...)

dpkoch: Okay cool, I'm good with that. I'll probably expose the clearing of the backup data in another merge request, but otherwise this is pretty great!

Issue #305, opened 2018-09-05
F4 Airspeed not working


sethholsinger:
I have been trying to get a openpilot revolution board to register a pixhawk airspeed sensor(JDrones). I have tried two boards and two i2c sensors both times the sensor did not register. I am using the flexiport (as it is the only one that seems to support i2c). Attached is my parameter file. Any ideas as to why this might be happening?

[F4_airspeed.txt](https://github.com/rosflight/firmware/files/2353522/F4_airspeed.txt)

superjax:
That is interesting.

One idea I have is to make sure that the board is powered with external 5V.

@BillThePlatypus, have you observed this?

BillThePlatypus:
I have not. I can check on it later today.

BillThePlatypus:
The power comment is really important. The usb port doesn't power the I2C.

sethholsinger:
I tried with and without power. Doesn't seem to register either way.

BillThePlatypus:
When you turn it on, do you leave the autopilot disarmed for a short time? It only looks for new sensors while the autopilot is disarmed.

sethholsinger:
I do not arm the autopilot as and just trying to test sensors.

sethholsinger:
Today a remade the cable that I am using to connect the i2c sensor to flexiport on the openpilot revo. And tried everything again and I am still not getting any airspeed data.

superjax:
That is kinda weird.  It is working on our end.

Could I recommend running the `airbourne` airspeed example?

`cd boards/airbourne/airbourne/examples/i2c_airspeed`
`make`
`make flash`
`miniterm.py /dev/ttyACM0`

You should be able to see the airspeed readings, and the yellow light (if you are using a revo) will go off.

superjax:
I also just tried it on the current `master` of `rosflight` and `master` on the `firmware`.  here is my setup.

![img_20180906_145911](https://user-images.githubusercontent.com/7344014/45184879-8615c180-b1e5-11e8-91bc-917afbfd5612.jpg)

dpkoch:
Have we verified that we're using the same sensor? The JDrones module uses a Measurement Specialties 4525DO, is that what we have? Also would it be worthwhile to run the `i2c_sniff` example from `airbourne_f4` to verify that the JDrones sensor shows up with the expected I2C address?

sethholsinger:
[...]

PR #335, opened 2019-02-07
Fix 305

len0rd:
Fixes for #305

1. Airspeed sensor is now recovered if there's intermittent external 5v power loss mid-flight (in this case the sensor must have been recognized before arming)
2. Airspeed sensor is now recognized when disarmed and if external 5v power is provided after boot up

For 2 above, I had to remove the [switch statement](https://github.com/rosflight/firmware/commit/eb269656205e2569a4330dc5c5fec8bd3333a88f#diff-d66352a6caa2830e83e1fd01f7619919L210) from sensors.look_for_disabled_sensors(). Functionally this means airspeed (and all other sensors) are checked once every second instead of once every 4 seconds. For whatever reason I could not get airspeed to pickup with the switch statement in place, even with increased check frequency (every 200ms instead of 1s). It was getting called in the switch. I'm willing to dig into this more if you feel this is an unsuitable fix.

How I reproduce the problem:
1. Plug in board via USB (no external power yet)
2. Start rosflight_io and verify there is no airspeed topic with `rostopic list`
3. Plug in external power. On current master, `rostopic list` will _not_ show /airspeed, this branch will

The CI build was also broken, so I updated the travis script to use xenial, and adjusted the packages accordingly. Builds now pass. Hope this is okay

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: "trusty"
+dist: "xenial"
 language: cpp
 
 addons:
@@ -6,9 +6,9 @@ addons:
     packages:
       - build-essential
       - libc6-i386
-      - lib32bz2-1.0
-      - lib32ncurses5
-      - lib32z1
+      - libbz2-1.0:i386
+      - libncurses5:i386
+      - libz1:i386
       - cmake
       - libgtest-dev
       - libeigen3-dev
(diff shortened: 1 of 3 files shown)


superjax:
Thanks, @len0rd!  This is great.

PR #336, opened 2019-02-12
Vscode debugging

len0rd:
Details on how I setup a firmware debug environment in Visual Studio code that works on both Mac and Linux (Ubuntu 18.04).

The master branch also has the material for MkDocs theme, so I don't think this can be easily applied to the currently live docs.


superjax:
Thanks man!  This is cool.

PR #337, opened 2019-03-23
Changed param callback method to listener interface

dpkoch:
Instead of each module registering callbacks for an individual param id, each module implements the `ParamListenerInterface` interface, and the param module calls their callback functions when a parameter is changed. This has some advantages and disadvantages over the current method:

## Advantages
  * No longer uses the heap; everything is done on the stack (main motivation)
  * Allows multiple modules to listen for changes to the same parameter. With the current implementation, if two modules tried registering callbacks for the same parameter, only the last callback registered would get called

## Disadvantages
  * Less efficient; before when a parameter was changed one callback was called, now every listener's callback is called. However, this seems like a reasonable tradeoff, especially since parameters aren't often changed while armed.


dpkoch:
Fixed the issue with the parameter initialization. The issue was that before every time a callback was registered, it got called immediately. I'm not doing that anymore, which means that every module is now responsible for doing a full initialization without relying on the callbacks, and only using the callback for applying changes that occur after initialization. `comm_manager` was the only module not doing that, but it's fixed now.

Issue #348, opened 2019-06-24
F4 External I2C Bug


superjax:
I just found a bug over the weekend.

TL;DR if the I2C tries to contact a device that isn't there, it throws a fit and won't talk any more.  That's why the onboard I2C works (those devices are always there) and why the external I2C seems buggy.

Long Version:
The STM32F4 I2C peripheral does not report an `ACK`.  It throws an interrupt on an `NACK`, but (probably for performance reasons), will let you continue on writing bytes to an absent device without any error indication even if the device isn't there.

I think this is the reason I had never seen a pure "polling" driver from STM.  All the blocking implementations are still interrupt-based, and I think it's because you need to be able to handle the `ACKNOWLEDGE_FAILURE` interrupt, which can come after you've already started sending bytes to the device.

Anyway, upon close inspection of the datasheet, it tells me that i have to make sure to send a `STOP` condition if I get one of these interrupts and reset the peripheral.   I'm not looking for the interrupt in my blocking `I2C::write` function so I end up leaving the peripheral in a bad state, and the `SDA` and `SCL` lines just remain low all the time, even though from a software perspective, things look like they are working.

This is a priority 1 bug, because a lot of people have been having poor performance from the airspeed. #140  #305.  I'll work on it as soon as I can, probably sometime this week.

plusk01:
Funny that this came up this weekend.

I also spent some time this weekend with the revo's I2C. I noticed the same thing about `AF` not being set until *after* you send some bytes on the bus.

https://github.com/plusk01/airdamon_f3/issues/11

At least I got to [use a `goto`](https://github.com/plusk01/airdamon_f3/blob/f4/chips/stm32f4xx/src/i2c.cpp#L94)...

superjax:
Nice!  I've never used `goto`. haha!

Yeah, I played with it a bit, and it looks like `AF` will get sent either by trying to send more bytes, or if you wait 25 milliseconds.  Which is like 100 years in microcontroller land...  I would really love a "got an ACK" status bit.  (Maybe I could read the GPIO line?)

plusk01:
Yeah, it seems to be working okay---I started bringing up I2C and testing with the baro ... but man, it is such a lame sensor to work with (the whole conversion and waiting for ADC before you read).

A couple of thoughts:
- `airdamon` I2C works, but is currently blocking.
[...]

Issue #140, opened 2017-04-26
Airspeed sensor error


srogers4:
![image](https://cloud.githubusercontent.com/assets/23121516/25447818/cc9a60b6-2a73-11e7-980b-494e4b706058.png)
Here is the plot from our test flight on 4/26/17. There are issues with the airspeed sensor regarding differential pressure.

superjax:
Sorry about this guys, this sensor is really a pain.

superjax:
Oh, I think I know what this is.

PR #349, opened 2019-06-27
New I2C Driver

superjax:
This is the new I2C driver.  It should be much better than the old one!  It seems to be a lot better at the hot-swap of sensors, so I think it will solve #348, #305 #140, which were all related to a bug in how `STOP` conditions were handled.

ToDo:

- [ ] extended hardware test on bench
- [ ] brown-out testing with airspeed
- [ ] Sonar and Airspeed hardware tests

Bonus!:
- [x] Get queue functionality of new driver working correctly (as noted in #348)


superjax:
Multi-i2c just failed after about 2 hours.  :(  I don't have time to fix it this week.

The new driver needs a better recovery scheme.  The `unstick` method doesn't seem to be working.

bsutherland333:
Closed due to inactivity.

PR #353, opened 2019-07-15
don't write beyond the end of esc_out_

willisj2:
I know this isn't into master - but gps is effectively operating as master right now.

This fixes the hardfault I was getting with the aux_command. Since `PWM_NUM_OUTPUTS` is the number of outputs the board supports, and the length of `esc_out_`, I think this is a reasonable way to fix this. Another option would be to reduce the size of the `aux_command` message to send only 11 values, but that would be more work. Happy to do it if you'd prefer that approach.


superjax:
Thanks @willisj2!  I think that this is great.

PR #356, opened 2019-07-17
fixed init_param bug to include null (strlen+1)

plusk01:
sometimes i hate cstrings

`strlen` does not count `\0`, so when `strlen()<PARAMS_NAME_LENGTH` the `+1` is needed so that `lookup_param_id` knows to stop.


superjax:
Really?  That is super weird...  I thought that `strlen` used `\0` to figure out the length?  You'd think it would count the `\0` because it's part of the string.

I agree with you.

PR #357, opened 2019-07-19
new error calculation for fusing external attitude

plusk01:
- [x] clean up param names
- [x] test in Gazebo SIL (closed loop)
- [x] test in Unity SIL (closed loop)
- [ ] test in hardware with mocap (closed loop)

---
### Overview

Mahony's nonlinear complementary filter:

<p align="center"></p>

Where `w_err` is formed using vector-based measurements. Many autopilot implementations of the Mahony filter (e.g., betaflight) use the accelerometer to calculate the angular rate correction term as

<p align="center"></p>

ROSflight find the shortest-path quaternion between gravity and the measure of gravity and then uses eq 47 of Mahony. This is an interesting discussion point in-and-of itself and some experiments lead me to believe that the betaflight way is *slightly* better, but the ROSflight way is much more satisfying. I can expand on this later if it's useful. I would love some help understanding this better.

To incorporate external attitude updates, we are currently doing `qhat boxminus qext`, which again is *so* satisfying. However, it seems to be causing problems as discussed and illustrated below. A method that uses vector measurements similar to what others do for accel / mag measurements (see also eq 32 of Mahony) is

<p align="center"></p>

where `Rbar` is `qext` and `R` is `qhat`. The `kext` is chosen to address the fact that ext att updates are integrated at a different rate than IMU.

I'm hoping you can help me understand *why* the current approach isn't working. It would be nice to keep/fix the current implementation. I tested a few different metrics for calculating attitude error as a vector (boxminus, axis-angle, error quaternion), but the vector method above showed the best performance.

---

### What I think is broken, how I fixed it, and some plots that might convince you

The current scheme for fusing external attitude breaks when the attitude goes through a 90 deg yaw. There is also an issue of handling the multirate integration (imu rate and ext att rate).

The proposed scheme uses the measurement vector approach by converting quaternions to rotmat and then taking cross products between measured and estimated vectors of the DCM.

The following plots are obtained using [rosflight_matlab](https://github.com/plusk01/rosflight_matlab) with matlab scripts [autopilot_estimators](https://github.com/plusk01/autopilot_estimators). Each dataset recorded IMU and mocap with no external attitude updates. The IMU is then fed to the ROSflight estimator to evaluate various schemes, etc.

For ease of comparison, code that makes `q.w` always positive if commented out in `Quaternion::normalize`.

---
### Gazebo SIL (IMU 250 Hz, ext att 100 Hz)

image

image

### MIT ACL Snapdragon autopilot in VICON flight space (IMU 200 Hz, ext att 100 Hz)

image

image

### Bag from https://github.com/rosflight/firmware/issues/289#issuecomment-418388412 (IMU 1000 Hz, ext att 15 Hz -- to simulate vision updates)

image

image


superjax:
Thank you so much for putting this together.  

This is tricky stuff.

I don't have time to look at this today, but I think it's really important.  I'll look at this hopefully tomorrow.

plusk01:
No worries @superjax, I know you're busy with writing.

I've been thinking about this off and on the past month so I figured I would start putting together what I have.

jakeware:
This is great.  We tested this in Unity with good results.  Thanks, @plusk01.

src/estimator.cpp
```diff
@@ -181,30 +188,49 @@ void Estimator::run()
...
-    w_acc.z = 0.0f;
+    // w_err = 2*s_tilde*v_tilde
+    w_err.x = -2.0f*q_tilde.w*q_tilde.x;
+    w_err.y = -2.0f*q_tilde.w*q_tilde.y;
+    w_err.z = 0.0f; // Don't correct z, because it's unobservable from the accelerometer
   }
 
   if (attitude_correction_next_run_)
   {
+    turbomath::Vector xhat_BW, yhat_BW, zhat_BW;
```
superjax: Thanks for working all this out.

What do you think about including Eigen as a dependency?  Perhaps that's out of the scope of this PR, but I think it would be nice to replace `turbomath::Quat` and `turbomath::Vector3` with Eigen.

superjax: Also, is there a chance you could write the associated latex code here as a comment?  It's a little too big to grok at once :)

src/estimator.cpp
```diff
@@ -181,30 +188,49 @@ void Estimator::run()
...
+    turbomath::Vector xext_BW, yext_BW, zext_BW;
+    {
+      float &w = state_.attitude.w, &x = state_.attitude.x, &y = state_.attitude.y, &z = state_.attitude.z;
+      xhat_BW.x = 1.0f - 2.0f*(y*y + z*z); xhat_BW.y = 2.0f*(x*y - z*w); xhat_BW.z = 2.0f*(x*z + y*w);
+      yhat_BW.x = 2.0f*(x*y + z*w); yhat_BW.y = 1.0f - 2.0f*(x*x + z*z); yhat_BW.z = 2.0f*(y*z - x*w);
+      zhat_BW.x = 2.0f*(x*z - y*w); zhat_BW.y = 2.0f*(y*z + x*w); zhat_BW.z = 1.0f - 2.0f*(x*x + y*y);
+    }
+    {
+      float &w = q_correction_.w, &x = q_correction_.x, &y = q_correction_.y, &z = q_correction_.z;
```
superjax: Does it make sense to make this a function?

I actually think this could be a good opportunity to break up `estimator::run` so it's a little more concise and delegate the nitty-gritties to other functions.

src/estimator.cpp
plusk01: I agree that using Eigen would be nice.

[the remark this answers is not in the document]
plusk01: world w.r.t body

It's inspired by [Drake notation](https://drake.mit.edu/doxygen_cxx/group__multibody__notation__basics.html).

plusk01:
sounds good. I'm adding some notes to the estimator write up so this is still wip, FYI.

superjax:
Cool.

dpkoch:
Yeah, it looks like the underlying mavlink submodule just hasn't been updated yet. This will require an updated version of `rosflight_io` as well

plusk01:
Needs https://github.com/rosflight/mavlink/pull/14 and https://github.com/rosflight/rosflight/pull/107.

Should be good to review now.

superjax:
Awesome.  I'm on vacation this week, @dpkoch, would you have time to test this out this week?

dpkoch:
Hi @plusk01, thanks for working on this. I'm unable to push changes to this branch, so could you update the mavlink submodule to the latest commit on master at https://github.com/rosflight/mavlink_c_library.git? Should just need to `cd comms/mavlink/v1.0`, `git checkout master`, `git pull`.

plusk01:
done!

dpkoch:
This branch is currently failing the exponential integration estimator tests; the errors are higher than the previous thresholds. Need to look into why that's changed and if it's a real issue

plusk01:
sorry I haven't followed up on this @dpkoch -- paper deadlines, etc. I should be able to take a look at it within the next week (if you haven't already).

dpkoch:
No worries @plusk01! I haven't done much in the last while either. With #361 merged, hopefully the unit tests will pass, although it will have introduced some merge conflicts that need to be resolved. We also need to uncomment the `MovingExtAtt` test in `test/estimator_test.cpp` and make sure that is passing now. If you have time to look at it that'd be great, but let me know if you'd like help.

superjax:
I also should have time to look at this this evening.

I can't spend too much time talking about it right now, but I have some suspicions on why the quaternion logic isn't working -- stay tuned.

superjax:
Hey, I want to revive this.  I've got a unit test in place that this stuff should make pass

plusk01:
superseded by #365

PR #361, opened 2019-08-18
Issue319 reorgnize unit tests

superjax:
I re-did all the unit tests.  :).  They needed some TLC.

Since we wrote these tests the first time, I have learned about test fixtures.  They are used pretty extensively in the new suite.

Everything that we used to test is still being tested, plus a few more things that weren't being tested.  I also expanded the estimator tests to include external attitude measurements.  This suite includes a unit test that shows the divergent behavior discussed in #357.  It doesn't pass on `master` so I commented it out, for now.

I also found _another_ bug in the external attitude measurement stuff that caused the gyro biases to not be estimated if accelerometer was off. Anyway, there was some reordering of logic in `estimator.c` to get this to work.  Unfortunately, it's probably going to cause conflicts when we merge #357.


dpkoch:
Reminder on #357 to uncomment the `MovingExtAtt` test in `estimator_test.cpp`

PR #365, opened 2019-10-19
Extatt

superjax:
This is a rebase of #357.


superjax:
The results of the unit tests are plotted with `scripts plot_estimator_tests.py`

Here are the results for `movingExtAtt` with this branch
image

And here are the results with `master`
image
scary.

superjax:
As for why this happened.  I have tracked it down to a bug in the `quaternion` `boxminus` logic.

There are some conditions where `q1 [-] q2` if `q1` and `q2` are really close, actually results in very large total errors (really close to `2pi`).

plusk01:
Once Travis finishes, https://github.com/rosflight/rosflight/pull/107 is good to merge as well

PR #366, opened 2019-11-12
Add MkDocs strict build to test script

dpkoch:
I've been setting up the webhook to host the documentation on our own server rather than GitHub pages, and thought I'd take the opportunity to improve the test script to ensure the documentation always builds. This PR changes two things:

  - Adds the mkdocs build to the test script (instead of just `.travis.yml` so the user gets feedback before pushing if there's an issue)
  - Runs the mkdocs build command in strict mode, so that it fails if there's a warning instead of ignoring things like missing images, bad links, etc.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,6 @@ before_script:
 
 script:
   - ./scripts/run_tests.sh
-  - mkdocs build
 
 deploy:
   provider: pages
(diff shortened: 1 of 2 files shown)


dpkoch:
I suppose the only downside is that now people who just want to work on code have to install the mkdocs dependencies to run the test script... but probably worth it?

superjax:
lgtm.  Thanks!

PR #371, opened 2020-01-23
Verbose RC Override

BillThePlatypus:
This changes the RC Override variable in the status message from a boolean to a bitfield, that indicates specifically why RC is overriding offboard controls. This also allows you to tell which channels are overridden. There are corresponding changes to rosflight_io that show this in the status messages.

### DIFF
--- a/include/command_manager.h
+++ b/include/command_manager.h
@@ -143,33 +143,66 @@ private:
     MUX_Z,
     MUX_F,
   };
+  enum RCOverrideReason: uint16_t
+  {
+    OVERRIDE_NO_OVERRIDE = 0x0,
+    OVERRIDE_ATT_SWITCH = 0x1,
+    OVERRIDE_THR_SWITCH = 0x2,
+    OVERRIDE_X = 0x4,
+    OVERRIDE_Y = 0x8,
+    OVERRIDE_Z = 0x10,
+    OVERRIDE_T = 0x20,
+    OVERRIDE_OFFBOARD_X_INACTIVE = 0x40,
+    OVERRIDE_OFFBOARD_Y_INACTIVE = 0x80,
+    OVERRIDE_OFFBOARD_Z_INACTIVE = 0x100,
+    OVERRIDE_OFFBOARD_T_INACTIVE = 0x200,
+  };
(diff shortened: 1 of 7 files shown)


include/command_manager.h
```diff
@@ -143,33 +143,66 @@ class CommandManager : public ParamListenerInterface
...
 
   typedef struct
   {
     RC::Stick rc_channel;
     uint32_t last_override_time;
-  } rc_stick_override_t;
+    RCOverrideReason stick_override_reason;
+    RCOverrideReason offboard_inactive_override_reason;
+    uint16_t override_mask;
+  } channel_override_t;
```
superjax: do we want to migrate to using C++ structs, instead of C structs?

src/command_manager.cpp
```diff
@@ -158,81 +158,56 @@ bool CommandManager::stick_deviated(MuxChannel channel)
...
+      rc_override |= channel_override_[channel].stick_override_reason;
+    if(!(muxes[channel].onboard->active))
+      rc_override |= channel_override_[channel].offboard_inactive_override_reason;
   }
-  // set the combined channel output depending on whether RC is overriding for this channel or not
-  *muxes[channel].combined = override_this_channel ? *muxes[channel].rc : *muxes[channel].onboard;
-  return override_this_channel;
+  if(!(muxes[MUX_F].onboard->active)) // The throttle has unique override behavior
+    rc_override |= OVERRIDE_OFFBOARD_T_INACTIVE;
+  if(RF_.params_.get_param_int(PARAM_RC_OVERRIDE_TAKE_MIN_THROTTLE))
```
superjax: I know that the logic is relatively obvious, but I got lost.  A comment would be nice.
also, let's put braces on this `if` statement

test/command_manager_test.cpp
```diff
@@ -47,10 +47,10 @@ class CommandManagerTest : public ::testing::Test
     rf.state_manager_.clear_error(rf.state_manager_.state().error_codes); // Clear All Errors to Start
     rf.params_.set_param_int(PARAM_CALIBRATE_GYRO_ON_ARM, false);
 
-    for (int i = 0; i < 8; i++)
-    {
+    for (int i = 0; i < 4; i++)
       rc_values[i] = 1500;
-    }
+    for (int i=4;i<8;i++)
```
superjax: formatting

bsutherland333:
This also seems like a lot of great work, we should visit revisit this at some point.

PR #377, opened 2020-02-07
Cleanup backup SRAM implementation

dpkoch:
This pull request cleans up the implementation of the backup SRAM hardfault handling. The main objectives were to remove the ROSflight dependencies from inside airbourne_f4, and to make it more clear in code when the backup memory is being read, erased, and written.

So far, this has been bench tested with the following results:

**Desired behaviors verified to be working:**

  - The firmware knows when it has recovered from a hardfault, and correctly reports the number of times it has rebooted from a hardfault since the last normal startup
  - The hardfault recovery information is deleted after rebooting; when restarted normally no hardfault recovery is detected
  - The arm state at hardfault is successfully restored on reboot. This arm state does not persist across a normal shutdown/startup

**Known issues:**
I'm not sure if the previous implementation had this issue or not; but either way it's a problem.

If you set the mixer and calibrate the IMU, but do _not_ do a param write, and if you then arm and encounter a hardfault, the system will restore the arm state, but will not retain the mixer setting and IMU calibration. So the system will be armed, but will have an invalid mixer and an uncalibrated IMU. Obviously this is undesirable.

Also, the arm status LED gets out of sync in this case; it blinks like it would if there were errors, and does not show the solid "armed" indicator

Options for fixing:

  1. Add additional logic to the bootup sequence that won't restore the armed state if conditions that would normally constitute and error are present. I need to look into how difficult this would be with the current boot sequence
  2. Make unwritten parameter values an error that prevents arming in the first place. This could be annoying for gain tuning, etc. Also since we currently don't disallow setting parameters while armed, this could have some weird conflicts.

I'm going to proceed with option 1, unless anyone can think of a reason to consider something else?


BillThePlatypus:
I hadn't thought of the unsaved mixer issues. Could other setting issues be a problem? I'm thinking `RC_TYPE` (even with the config manager), `ARM_CHANNEL` (mostly because it would prevent disarming, or could result in an immediate disarm if mapped to the wrong switch). `MOTOR_PWM_UPDATE` and `AIL_REV` and the like could also cause crashes, depending.

dpkoch:
Okay, I think I have it fixed now.

The way it was set up (before this PR, and then I mirrored that implementation) was that the firmware would rearm as almost the first thing. However, I think this was really dangerous, because almost nothing had been initialized yet (params, RC, mixer, etc.). I suppose maybe it potentially wasn't too terrible since everything would still initialize before any loops run, but still. Not ideal.

I changed it so that the firmware checks the backup memory as the _last_ step in its initialization, not the first. That way everything is already up and running, and now we can only rearm if we're free of errors. This fixes the issue where the e.g. the mixer was not saved, but we were arming anyway. Now if there's an error condition it will simply not rearm, and send a message to the user.

Obviously this introduces its own problems. As long as parameters are written before arming, it should reinitialize just fine. So do we just put in the documentation that you should always write params before arming? Or do we enforce that in code somehow?

BillThePlatypus:
Are there any cases where someone would prefer not to save parameters
before arming? Saving parameters is quick.

BillThePlatypus:
Could flight-essential params be saved in case of a hard fault? That may add too much complexity, because in case of a hard fault, we want to do relatively little, because something is broken.

include/util.h
```diff
@@ -0,0 +1,82 @@
...
+  {
+    block_length = len < max_block_length ? len : max_block_length;
+    for (size_t i = 0; i < block_length; i++)
+    {
+      c1 += *(src++);
+      c2 += c1;
+    }
+
+    c1 %= 255;
+    c2 %= 255;
```
BillThePlatypus: Using mod 255 means that 0xff is identical to 0x00. Could we use a different modulus, perhaps 253? As I'm seeing it, if the data is filled with 1's, it will be treated as valid. It would just send a nuisance crash message, though.

test/test_board.h
```diff
@@ -122,9 +122,10 @@ class testBoard : public Board
   void led1_toggle() override;
 
 //Backup memory
-  bool has_backup_data() override;
-  BackupData get_backup_data() override;
-
+  void backup_memory_init() override {}
+  bool backup_memory_read(void *dest, size_t len) override { (void)dest; (void)len; return false; }
+  void backup_memory_write(const void *src, size_t len) override { (void)src; (void)len; }
+  void backup_memory_clear(size_t len) override { (void)len; }
```
BillThePlatypus: It might be good to add some tests for such things as not accidentally rearming, but that could perhaps wait for a future PR.

include/util.h
```diff
@@ -0,0 +1,82 @@
+/*
+ * Copyright (c) 2019, James Jackson and Daniel Koch, BYU MAGICC Lab
```
superjax: We should add @BillThePlatypus to this file, for sure

include/util.h
superjax: would 256 work?

BillThePlatypus: I think it is intentional that it does not line up with the size of a byte.

BillThePlatypus:
This works well in simulation.
It works in hardware (no props yet), but only if USB is plugged in.

BillThePlatypus:
It looks like one potential problem is that it will not rearm if there are any errors. Notably we can expect there to be an IMU uncalibrated error, and an RC lost error is possible (In which case we would want to go to a failsafe command). As is, this would prevent rearming if the error is processed before backup data.

test/test_board.h
superjax: I think that's a pretty important thing to do.

#381

dpkoch:
Also need to discuss in the documentation

#382

PR #378, opened 2020-02-14
Battery monitor

### LINKED ISSUE #144: Battery Info
This would be awesome, published on status message

BillThePlatypus:
Support for analog battery monitors, including both current and voltage
This adds the `BATT_VOLT_MULT`, `BATT_CURR_MULT` and `STRM_BATTERY` parameters.

Closes #144


docs/user-guide/hardware-setup.md
```diff
@@ -98,6 +99,13 @@ You will need a laptop which can run Ubuntu 16.04 or 18.04 with ROS to communica
...
+
+For ROSflight to use a battery monitor, an appropriate multiplier must be set. ROSflight multiplies the analog signal from the monitor by the multiplier to get the final reading. Themonitor datasheet should contain the information needed to get the multiplier. For example, the datasheet for the AttoPilot 50V/90A sensor states that it outputs 63.69 mV / V. To get the original battery voltage, the multiplier must be 1/.06369, or 15.7. The multipliers for the voltage and current are set separately, with the `BATT_VOLT_MULT` and `BATT_CURR_MULT` parameters, respectively. In addition, the `STRM_BATTERY` parameter must be set in order to stream the battery information.
```
superjax: Typo themonitor

boards/airbourne/Makefile
```diff
@@ -128,7 +128,10 @@ AIRBOURNE_SRCS = led.cpp \
                  uart.cpp \
                  M25P16.cpp \
                  ms4525.cpp \
-                 backup_sram.cpp
+                 backup_sram.cpp \
+                 analog_digital_converter.cpp \
+		 analog_pin.cpp \
```
superjax: [nit] spacing

boards/airbourne/airbourne_board.cpp
```diff
@@ -145,6 +145,8 @@ void AirbourneBoard::sensors_init()
   sonar_.init(&ext_i2c_);
   airspeed_.init(&ext_i2c_);
   // gnss_.init(&uart1_);
+  battery_adc_.init(&adc_config[0]);
```
superjax: can we replace the `0` with a named value?

src/sensors.cpp
```diff
@@ -209,6 +216,28 @@ void Sensors::update_other_sensors()
...
+            battery_montitor_filter_pointer = 0;
+          float voltage_sum{0};
+          for(float voltage: battery_voltage_history)
+            voltage_sum += voltage;
+          data_.battery_voltage = voltage_sum / BATTERY_MONITOR_MOVING_AVERAGE_COUNT;
+        }
+        if(rf_.board_.battery_current_present())
+        {
+          data_.battery_monitor_present = true;
+          data_.battery_current = rf_.board_.battery_current_read();
```
superjax: There is a lot of logic here.

Could we move this to another function?  (Is there a reason we can't use the outlier_filter?)

BillThePlatypus: I'm moving to another function. The purpose here is not to remove outliers, but to filter the pretty severe noise. The monitor should really have a hardware filter.

boards/airbourne/airbourne_board.cpp
```diff
@@ -331,6 +333,36 @@ GNSSRaw AirbourneBoard::gnss_raw_read()
...
+{
+  return this->battery_monitor_.has_voltage_sense();
+}
+
+float AirbourneBoard::battery_voltage_read() const
+{
+  return static_cast<float>(this->battery_monitor_.read_voltage());
+}
+
+void AirbourneBoard::battery_voltage_set_multiplier(double multiplier)
```
dpkoch: Does this multiplier need to happen in the board layer? Should this be managed in the firmware sensors module?

include/sensors.h
```diff
@@ -201,6 +205,8 @@ class Sensors : public ParamListenerInterface
   static const int SENSOR_CAL_CYCLES;
   static const float BARO_MAX_CALIBRATION_VARIANCE;
   static const float DIFF_PRESSURE_MAX_CALIBRATION_VARIANCE;
+  static constexpr size_t BATTERY_MONITOR_MOVING_AVERAGE_COUNT = 10;
+  static constexpr uint32_t BATTERY_MONITOR_UPDATE_PERIOD = 100;
```
dpkoch: Can we add units to this constant name (e.g. prepend with `_MS` or `_US`)?

src/sensors.cpp
```diff
@@ -339,6 +362,16 @@ bool Sensors::update_imu(void)
...
 }
 
+void Sensors::update_battery_average(float new_voltage)
+{
+  battery_voltage_history_[battery_monitor_filter_pointer_++]=new_voltage;
+  if(battery_monitor_filter_pointer_ == BATTERY_MONITOR_MOVING_AVERAGE_COUNT)
+    battery_monitor_filter_pointer_ = 0;
+  float voltage_sum{0};
+  for(float voltage: battery_voltage_history_)
+    voltage_sum += voltage;
```
dpkoch: It should be possible to implement this without having to sum over the whole array every time, right?

Like
```cpp
data_.battery_voltage += (battery_voltage_history[newest_index] - battery_voltage_history_[oldest_index]) / BATTERY_MONITOR_MOVING_AVERAGE_COUNT;
```
Is that worth doing to save a few cycles? Or does it just not matter?

src/sensors.cpp
```diff
@@ -642,4 +675,12 @@ bool Sensors::OutlierFilter::update(float new_val, float *val)
   }
 }
 
+void Sensors::update_battery_monitor_multipliers()
+{
+  float voltage_multiplier = this->rf_.params_.get_param_float(PARAM_BATTERY_VOLTAGE_MULTIPLIER);
+  float current_multiplier = this->rf_.params_.get_param_float(PARAM_BATTERY_CURRENT_MULTIPLIER);
+  this->rf_.board_.battery_voltage_set_multiplier(voltage_multiplier);
+  this->rf_.board_.battery_current_set_multiplier(current_multiplier);
```
dpkoch: Again, would it be better to just store this multiplier in the sensors module (or just get the parameter value every time), rather than pass it all the way down to airbourne driver?

PR #379, opened 2020-02-28
fix RC Raw streaming

### LINKED ISSUE #370: Fix reporting RC values in the 0 to 1000 range
Right now we're reporting RC values (as published on `rc_raw`) in the 0 to 1000 range, instead of in the 1000 to 2000 range. This has been causing people confusion.

It may be a rosflight_io thing, but off the top of my head I think it was a change to the firmware that made the change.

We should either revert to 1000 to 2000, or (maybe better?) report floating point from 0 to 1 or -1 to 1 or something. Do you have thoughts @superjax?

We also need to make sure the documentation is consistent with what we're reporting.

superjax:
Fixes #370


BillThePlatypus:
Do we know of any packages that may depend on reading these incorrect values from `RC_RAW`?

superjax:
I can't think of any (famous last words)

PR #384, opened 2020-03-12
Auto release

superjax:
This PR adds two github actions.  
- The first releases the binaries (`.elf` and `.bin`) for both F1 and F4 targets for every push to `master`.  This is labeled as a `Pre-release` and named `Development Build`
 - The second automatically builds and releases binaries and source code for every git tag labeled according to `v1.3.0` or something like that.

After this PR is merged:
 - to update the current "pre-release" Development build, merge with `master`
 - to create a new release, `git tag` and `git push --tags origin`.

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -1,0 +1,43 @@
+---
+name: "pre-release"
+
+on:
+  push:
+    branches:
+      - "master"
+
+jobs:
+  pre-release:
+    name: "Pre Release"
+    runs-on: "ubuntu-latest"
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: clone
+      run: git submodule update --init --recursive
+    - name: install toolchain
+      run: curl --retry 10 --retry-max-time 120 -L "https://launchpad.net/gcc-arm-embedded/5.0/5-2016-q1-update/+download/gcc-arm-none-eabi-5_3-2016q1-20160330-linux.tar.bz2" | tar xfj -
+    - name: check toolchain
+      run: |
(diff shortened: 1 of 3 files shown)


comms/mavlink/mavlink.h
```diff
@@ -37,6 +37,8 @@
 #pragma GCC diagnostic ignored "-Wswitch-default"
 #pragma GCC diagnostic ignored "-Wcast-align"
 #pragma GCC diagnostic ignored "-Wignored-qualifiers"
+#pragma GCC diagnostic ignored "-Wpragmas"
+#pragma GCC diagnostic ignored "-Waddress-of-packed-member"
```
superjax: This updates us to GCC 9

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,43 @@
...
+  pre-release:
+    name: "Pre Release"
+    runs-on: "ubuntu-latest"
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: clone
+      run: git submodule update --init --recursive
+    - name: install toolchain
+      run: curl --retry 10 --retry-max-time 120 -L "https://launchpad.net/gcc-arm-embedded/5.0/5-2016-q1-update/+download/gcc-arm-none-eabi-5_3-2016q1-20160330-linux.tar.bz2" | tar xfj -
```
dpkoch: Isn't this really outdated? The docs say we're targeting 7.3.1, and that's what we get using the PPA (also what's in the developer docs). Should we switch to a PPA install here? Might also need to update the CI actions

dpkoch:
I just opened #385 which updates the CI workflows to install arm GCC from the PPA (it gets 7.3.1). We could update this to match

PR #386, opened 2020-03-13
Tweaks to automatic release/latest builds

dpkoch:
Changes:

## Change prerelease tag
Use "latest" instead of "pre-release"

Rationale: I think "pre-release" is misleading, since it has the implication that it's a release candidate build. I think latest conveys the meaning more clearly. And since it's also titled "Development build" and marked as a pre-release, I think it's pretty clear.

Also, this is the tag that will get used in the compiled-in version tag. I think that it makes more sense to get `latest-0-g1234abc` than `pre-release-0-g1234abc`.

## Don't attach .elf files

As far as I know we don't actually use these, right? So I think it avoids a point of confusion for people who are just trying to find the file they need to flash. They only have to choose Naze vs. Revo, not `.hex`/`.bin` vs. `.elf`

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -13,7 +13,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: clone
+    - name: checkout submodules
       run: git submodule update --init --recursive
     - name: install toolchain
       run: |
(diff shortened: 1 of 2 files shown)


superjax:
That sounds reasonable!

I guess if you're going to debug (the reason you need the `.elf`), then you'd be building from source anyways

PR #393, opened 2020-04-17
Gnss reenable

BillThePlatypus:
Reenables GNSS support on the main port for the revo.

### DIFF
--- a/scripts/fix-code-style.sh
+++ b/scripts/fix-code-style.sh
@@ -7,5 +7,5 @@ echo #SCRIPTPATH
 cd $SCRIPTPATH/..
 
 find . -iname "*.h" -o -iname "*.cpp" | \
-egrep -v "^(./comms/mavlink/v1.0/|./board/airbourne/airbourne/lib/|./boards/breezy/breezystm32|./.git|./test/build)" | \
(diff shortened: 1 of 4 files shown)


scripts/fix-code-style.sh
```diff
@@ -7,5 +7,5 @@ echo #SCRIPTPATH
 cd $SCRIPTPATH/..
 
 find . -iname "*.h" -o -iname "*.cpp" | \
-egrep -v "^(./comms/mavlink/v1.0/|./boards/airbourne/airbourne/|./boards/breezy/breezystm32|./.git|./test/build)" | \
+egrep -v "^(./comms/mavlink/v1.0/|./boards/airbourne/airbourne/lib/|./boards/breezy/breezystm32|./.git|./test/build)" | \
```
dpkoch, line 10: Shouldn't we ignore the whole airbourne directory, since it's a submodule and it's formatting is handled independently?

BillThePlatypus: There isn't any formatting information in airbourne. Additionally, IDE's don't necessarily differentiate between firmware and airbourne.

That would be cleaner. This was mostly a quick fix so that PR's wouldn't have a ton of whitespace changes.

PR #394, opened 2020-04-23
Update gnss fix types, actually send them

BillThePlatypus:
Sends actually fix information for GNSS. The firmware now has its own enum type for fix type, because the ROS and UBLOX fix types don't fit our needs well.


BillThePlatypus:
Mavlink and Airbourne PRs are pending.

BillThePlatypus:
This is now ready for review.

PR #395, opened 2020-04-29
Require users to set the failsafe throttle

BillThePlatypus:
This changes the default failsafe throttle to -1.0, and refuses to arm if it is not set to a valid value (0.0-1.0). This will hopefully make multirotor flight safer.
For fixed-wing aircraft, this is ignored, as they always go to zero throttle in case of RC failure.


src/command_manager.cpp
```diff
@@ -82,9 +82,22 @@ void CommandManager::param_change_callback(uint16_t param_id)
 
 void CommandManager::init_failsafe()
 {
-  multirotor_failsafe_command_.F.value = RF_.params_.get_param_float(PARAM_FAILSAFE_THROTTLE);
+  float failsafe_thr_param = RF_.params_.get_param_float(PARAM_FAILSAFE_THROTTLE);
+  bool fixedwing = RF_.params_.get_param_int(PARAM_FIXED_WING);
+  if (!fixedwing
+      && (failsafe_thr_param < 0. || failsafe_thr_param > 1.0)) // fixed wings always have 0 failsafe throttle
```
superjax, line 88: I don't think we need to enforce this.  Our mantra so far has been "we are all adults here", so I'm fine if someone wants a non-zero failsafe throttle if they're using a fixedwing

BillThePlatypus: No, this allows for the failsafe not to be configured for fixedwings, because it is ignored.

Issue #148, opened 2017-05-03
Support Arbitrary IMU Orientation


superjax:
I had to mount my naze for my autogyro upside down.  How do I fly?

TimWhiting:
Wondering what the status of this issue is, because our team has the naze upside down on the foamy... Can I help get this working?

TimWhiting:
I looked into the branch for this fix. It looks like you were planning on supporting any arbitrary angle for the orientation. Is that necessary or could we just multiply the raw imu input by -1 based on an orientation parameter in the x y and z axis?

superjax:
Sort of. You have to multiply both the z and y axes by -1, (otherwise you'll end up with a left-handed coordinate frame).  Since there are so many configurations to consider, it just made sense to support arbitrary orientations.  There is actuallly a pull request active on this one, which adds 3 parameters.  If you merged `master` into that branch, it would probably work.

legoabram:
I had to make a small tweak, but I have it working now. The parameters say degrees, but everything is treated like radians. I figured degrees made more sense in this context, so I added a scalar. I've also made a few organization changes to match what I've been seeing elsewhere in the code. If you would allow it, how may I submit my changes?

superjax:
Why don't you submit a pull request and put me as a reviewer?

superjax:
Finished!

PR #408, opened 2023-09-08
Added udp_board.cpp files from ROS rosflight_firmware package

bsutherland333:
As part of [rosflight pull request #148](https://github.com/rosflight/rosflight/pull/148), I'm moving the udp board implementation directly into the rosflight_firmware package since the udp board doesn't need ROS and fits better directly in the firmware.

### DIFF
--- a/boards/udp/udp_board.cpp
+++ b/boards/udp/udp_board.cpp
@@ -34,18 +34,24 @@
  * \author Daniel Koch
  */
 
-#include <iostream>
 #include "udp_board.hpp"
+
+#include <iostream>
 
 using boost::asio::ip::udp;
 
 namespace rosflight_firmware
 {
-UDPBoard::UDPBoard(std::string bind_host, uint16_t bind_port, std::string remote_host,
-                   uint16_t remote_port)
-    : bind_host_(std::move(bind_host)), bind_port_(bind_port), remote_host_(std::move(remote_host)),
-      remote_port_(remote_port), io_service_(), socket_(io_service_), write_in_progress_(false)
-{}
+UDPBoard::UDPBoard(std::string bind_host, uint16_t bind_port, std::string remote_host, uint16_t remote_port) :
+  bind_host_(std::move(bind_host)),
+  bind_port_(bind_port),
+  remote_host_(std::move(remote_host)),
+  remote_port_(remote_port),
+  io_service_(),
+  socket_(io_service_),
+  write_in_progress_(false)
+{
+}
 
 UDPBoard::~UDPBoard()
 {
@@ -55,11 +61,13 @@ UDPBoard::~UDPBoard()
   io_service_.stop();
   socket_.close();
 
-  if (io_thread_.joinable()) { io_thread_.join(); }
+  if (io_thread_.joinable())
+  {
+    io_thread_.join();
+  }
 }
 
-void UDPBoard::set_ports(std::string bind_host, uint16_t bind_port, std::string remote_host,
-                         uint16_t remote_port)
+void UDPBoard::set_ports(std::string bind_host, uint16_t bind_port, std::string remote_host, uint16_t remote_port)
 {
   bind_host_ = std::move(bind_host);
   bind_port_ = bind_port;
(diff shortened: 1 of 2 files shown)


dpkoch (commented):
I understand the motivation for this change, but it's somewhat awkward because it separates the source files for `udp_board` from its build files. And we definitely don't want to pull all of the boost dependencies into the firmware, so I don't think we want to bring any build files over. I'm open to suggestions for resolving those issues, but otherwise I would recommend against this change

bsutherland333:
That is a good point. I think my initial thought was that if udp_board is intended to be a general use case UDP implementation of the ROSflight firmware, then it would fit better within the ROSflight firmware itself with the other board implementations. I also figured that when we update the build files for the firmware to use cmake we could add a target for the udp_board pretty easily. But that still would require boost. Perhaps we could make it an optional dependency? 

Maybe the question that should be asked first is would we expect this to be a general use udp board for anyone who wants to run the firmware not on a microcontroller (like in a computer-only setup), or do we expect this to be used only for sim and SIL? If we only expect this to be used for SIL purposes then we could move it to the rosflight_sim package, next to the rosflight_sil board.

If putting it in the rosflight_firmware repo or the sim package both seem like a bad idea, then I could move it back into the rosflight_firmware package. My initial motivation for considering the move was to make the naming of everything more clear, by making the rosflight/rosflight_firmware package contain only the ROS build files for the rosflight_firmware repo, so it would be mostly "one in the same", with the ROS side only having the files required to build it in a ROS framework. If there were significant differences between the two (like one containing board layers the other doesn't have) then them both being named the same thing would be confusing. Fortunately, an easy solution to that would be to just rename the rosflight/rosflight_firmware package to something like rosflight_udp_firmware, rosflight_udp_board, or rosflight_ros_firmware. 

I should also mention that everyone who's been working on the project recently agree that a name change is necessary. Simply calling the firmware repo `firmware` while the rosflight_io package is named `rosflight` under a repo named `rosflight` which doesn't actually contain the primary `rosflight` product (the firmware) has consistently confused everyone who wasn't already familiar with the project, and lead to some confusion during communication. So we decided that calling the whole thing rosflight and the individual parts `rosflight_io` and `rosflight_firmware` would be more clear.

dpkoch:
The renaming sounds good; I'm on board with that.

The original intent of the udp board was purely for SIL/sim, though we had put it in the `rosflight/rosflight_firmware` location so somebody could build a sim on top of it without depending on the whole gazebo sim implementation in `rosflight/rosflight_sim`. The naming issue you bring up makes sense. Moving it to a standalone library within `rosflight_sim` package would probably accomplish something similar, or a `rosflight_udp_board` package would be a bit more generic (though maybe overkill).

That's an interesting idea though about making it available as a starting point for computer-only setups. In that case it seems okay to move it into the firmware; I'd again just be very careful to keep those boost dependencies out of any core library or embedded targets. Making sure the dependency is optional (so embedded targets can be build without having to have boost available) would also be good. The separation of build files and source files is still a little awkward, but probably no worse than the existing ROS build files being kept in another repository (another casualty of the probably ill-advised decision to keep the firmware and ROS packages in separate repos).

TL;DR with that additional context, this seems reasonable to me, though some caution is needed. Sorry for slowing things down

bsutherland333:
Although we've toyed around with the idea of a computer only implementation, I think we decided that it would be easiest to still use a micro-controller running some stripped-down firmware as the amount of devices you'd need to plug into the computer to connect all the peripheral would be more complicated than using an existing flight controller as a sensor and actuator aggregator. So perhaps we can assume that nobody is going to need udp_board except for SIL/simulation stuff and put it next to the SIL board implementation.

On that note, we also plan to separate the simulation from Gazebo to make it more modular, where the simulation would run in standalone code and various visualization/collision tools could be plugged into it, like Gazebo or a high fidelity sim. So, we'll likely restructure the rosflight_sim package to be less tied to large dependencies anyways, avoiding the problem of including unneeded dependencies to use udp_board.

I'll probably just do that then, modifying the rosflight repo to reflect the changes.

dpkoch:
Cool, sounds good. Nice, I really like the plan to separate the simulation core from Gazebo.

PR #410, opened 2023-11-15
Reworked sensor and RC acquisition

bsutherland333:
Previously, sensors and RC hardware would collect and store data at a rate that was independent of how often the core rosflight code checked for sensor data. This commit changes that, making the core rosflight code be dependent on the hardware's aquisition rate, syncronizing the two.


bsutherland333:
This will be easier to push in with the varmint firmware, since this breaks the f1 and f4 firmware.

bsutherland333:
This pull request breaks the F1 and F4 firmware builds, but these builds will be replaced immediately by the [H7 varmint pull request](https://github.com/rosflight/rosflight_firmware/pull/411
).

JMoore5353 (approved):

PR #411, opened 2023-12-08
Varmint board

bsutherland333:
The existing F1 and F4 boards have long since gone out of production, and can no longer be bought. As part of our v2 update effort, we are adding support for new STM32H7 based boards. This board, the Varmint flight controller, is a custom board developed by AeroVironment that we've been using for our development and testing efforts. It's planned to be available commercially sometime next year. 

This board also uses the exact same STM processor as the CubePilot Orange (not Orange+) which we intend to add support for in our v2 release.


bsutherland333:
Since the ST provided code for this board implementation is 275,000 lines, I'm probably going to include this as a sub-module instead of being directly in the firmware repo and delete this branch so it doesn't clog up the git history with 27500 lines of code...

PR #412, opened 2023-12-15
Updated linter and .clang-format for ROS2

bsutherland333:
Some of the style guidelines have changed for ROS2, and since we want to keep a consistent style throughout ROSflight, I'm updating the firmware as well. 

Lately, the existing format linter has also been disagreeing with the changes made by the fix_code_style.sh script, which has been pretty annoying. So this updates the action item to use a newer version as well, fixing the issue.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,53 +1,68 @@
----
-Language:      Cpp
-BasedOnStyle:  Google
-IndentWidth: 2
-UseTab: Never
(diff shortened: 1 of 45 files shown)


JMoore5353 (approved):

bsutherland333:
F1 and F4 builds are currently broken, but are going to be removed shortly and replaced with a working build.

PR #413, opened 2024-01-03
Varmint submodule

bsutherland333:
This pull request adds support for AeroVironment's Varmint FCU package. It is being included as a submodule to keep this repo from getting cluttered with all of ST's drivers. It uses the work that was being developed in pull request [#411](https://github.com/rosflight/rosflight_firmware/pull/411).


bsutherland333:
This pull request does a few things:
- Removes boards Airborne and Breezy, replacing them with Varmint.
- Updates project to use CMake rather than Makefiles.
- Adds a quality of service argument to serial_write. AeroVironment added QOS to the Varmint to improve consistency of serial communication, but to avoid adding MAVlink as a dependency to the Varmint board code, I moved declaring the QOS level to the core firmware.
- Moved the main function from being part of the board implementation to the core firmware. Previously, a different main function was written for every board and was kept separate from the board's driver repository. To avoid needing a different main for every board and to minimize separation of board code with that board's repository, I made the the main function more general, with the only difference between each board being what board is being included and instantiated.

iandareid (approved):

PR #426, opened 2024-07-01
Update to support per PWM output rate, vector servo update, and DSHOT…

avtoku:
> Per PWM output rate 
> Single call to update all PWM's
> DSHOT frequencies (and protocol)
> Also fixed a missing check for 'M' servo type.
> Increased number of possible mixer servos from 8 to 10

### DIFF
--- a/include/board.h
+++ b/include/board.h
@@ -116,8 +116,10 @@ public:
 
   // PWM
   virtual void pwm_init(uint32_t refresh_rate, uint16_t idle_pwm) = 0;
+  virtual void pwm_init(const float *rate, uint32_t channels) = 0; // PTT new
   virtual void pwm_disable() = 0;
   virtual void pwm_write(uint8_t channel, float value) = 0;
+  virtual void pwm_write(float *value, uint32_t channels) = 0; //PTT new
 
   // non-volatile memory
   virtual void memory_init() = 0;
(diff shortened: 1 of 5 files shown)


bsutherland333:
I don't quite understand everything you're doing, could you explain your changes more?

avtoku:
Servo Frequency Update:
The previous version of PWM and MIXER:
1) Did not support DSHOT-type servo/motor channels
2)  Specified a single period for all pwm whether they were motors operating at 490Hz or Servos at 50Hz.
3) Updated the channel PWM's in independent calls to the board.
4) Channels in the mixer were updated from 8 to 10 because Varmint has 10 PWM outputs.
The new version
1) Supports DSHOT protocol 
2) Allows specification of different frequencies for each servo/motor channel (hardware permitting*). If the rate is specified in the range  150000 to 1200000 (bps), it is a DSHOT channel, and 50 to 490 (Hz)_ if it is a PWM channel. 
3) In DSHOT mode, all the channels within a group must be updated simultaneously if very high update update rates are to be achieved.

*On our processors, up to 4 PWM/DSHOT output channels can be assigned to a Timer. All 4 channels must be of the same rate. On the Varmint, PWM Timer groups are 1-4, 5-8, & 9-10, on PixRacer Pro they are 1-4, 5-6, & 7-8.

bsutherland333:
For now, I set all fixedwing motors to 50hz pwm and multirotor motors to 490hz, with all servos at 50hz. Eventually we will want to revamp the mixing system to make it easy to specify which motors to use dshot and which to use PWM. For now, using DSHOT will require modifying the source code.

PR #427, opened 2024-07-09
Removed outlier filters and artifacts

avtoku:
1. Removed outlier filters. This affects the airspeed, barometer, and sonar.  
2. Removed the (redundant) check for valid data in the comm_manager because those functions are not called unless the data is valid. This affects all send_* functions that have this redundancy.

### DIFF
--- a/include/sensors.h
+++ b/include/sensors.h
@@ -149,12 +149,10 @@ public:
     float diff_pressure_velocity = 0;
     float diff_pressure = 0;
     float diff_pressure_temp = 0;
-  //  bool diff_pressure_valid = false;
 
     float baro_altitude = 0;
     float baro_pressure = 0;
     float baro_temperature = 0;
-  //  bool baro_valid = false;
 
     float sonar_range = 0;
     bool sonar_range_valid = false;
--- a/include/sensors.h
+++ b/include/sensors.h
@@ -204,17 +204,10 @@ public:
   got_flags got;
 
 private:
-  static const float BARO_MAX_CHANGE_RATE;
-  static const float BARO_SAMPLE_RATE;
-  static const float DIFF_MAX_CHANGE_RATE;
-  static const float DIFF_SAMPLE_RATE;
-  static const float SONAR_MAX_CHANGE_RATE;
-  static const float SONAR_SAMPLE_RATE;
   static const int SENSOR_CAL_DELAY_CYCLES;
   static const int SENSOR_CAL_CYCLES;
   static const float BARO_MAX_CALIBRATION_VARIANCE;
   static const float DIFF_PRESSURE_MAX_CALIBRATION_VARIANCE;
-  static constexpr uint32_t BATTERY_MONITOR_UPDATE_PERIOD_MS = 10;
 
   enum : uint8_t
   {
--- a/include/sensors.h
+++ b/include/sensors.h
@@ -149,12 +149,12 @@ public:
     float diff_pressure_velocity = 0;
     float diff_pressure = 0;
     float diff_pressure_temp = 0;
-    bool diff_pressure_valid = false;
+  //  bool diff_pressure_valid = false;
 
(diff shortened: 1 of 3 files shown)


include/sensors.h
```diff
@@ -149,12 +149,12 @@ class Sensors : public ParamListenerInterface
     float diff_pressure_velocity = 0;
     float diff_pressure = 0;
     float diff_pressure_temp = 0;
-    bool diff_pressure_valid = false;
+  //  bool diff_pressure_valid = false;
```
bsutherland333, line 152: Could you remove this line instead of just comment it out?

include/sensors.h
```diff
@@ -149,12 +149,12 @@ class Sensors : public ParamListenerInterface
...
 
     float baro_altitude = 0;
     float baro_pressure = 0;
     float baro_temperature = 0;
-    bool baro_valid = false;
+  //  bool baro_valid = false;
```
bsutherland333, line 157: Same as above

src/comm_manager.cpp
```diff
@@ -397,37 +397,30 @@ void CommManager::send_rc_raw(void)
...
     comm_link_.send_sonar(sysid_,
                           0, // TODO set sensor type (sonar/lidar), use enum
                           RF_.sensors_.data().sonar_range, 8.0, 0.25);
   }
-}
 
 void CommManager::send_mag(void)
 {
-  if (RF_.sensors_.data().mag_present) comm_link_.send_mag(sysid_, RF_.sensors_.data().mag);
+ 	  	  comm_link_.send_mag(sysid_, RF_.sensors_.data().mag);
```
bsutherland333, line 420: This has tabs instead of spaces. The ROS style guide directs to use spaces instead of tabs, and mixing them makes for very messy formatting. clang-format can fix this for you if you'd like.

src/comm_manager.cpp
```diff
@@ -397,37 +397,30 @@ void CommManager::send_rc_raw(void)
 
 void CommManager::send_diff_pressure(void)
 {
-  if (RF_.sensors_.data().diff_pressure_valid) {
     comm_link_.send_diff_pressure(sysid_, RF_.sensors_.data().diff_pressure_velocity,
                                   RF_.sensors_.data().diff_pressure,
                                   RF_.sensors_.data().diff_pressure_temp);
   }
```
bsutherland333, line 403: I think you have some misaligned brackets and spacing in this file

bsutherland333 (changes requested):
Just a few formatting comments. If you run the clang-format script that uses the .clang-format file in the rosflight_firmware repository, then it should fix the spacing errors for you.

bsutherland333:
One other question, you mention that these functions are not called unless the data is valid so check if the data is valid is redundant. Where is the other validity check specifically? Is that in the rosflight_firmware code or the board implementation code?

include/sensors.h
avtoku: updated.

include/sensors.h
avtoku: updated

src/comm_manager.cpp
avtoku: ran clang-tools on that file.

avtoku:
> One other question, you mention that these functions are not called unless the data is valid so check if the data is valid is redundant. Where is the other validity check specifically? Is that in the rosflight_firmware code or the board implementation code?

Most of the checks that were removed were of the form "if(RF_.sensors_.data().mag_present)" and has been replaced with the   if "if(got.mag) send_mag();" check in CommManager::stream(got_flags got).

Baro, Pitot, and Sonar were also blocked from sending if there were outliers which we are not checking anymore. 

In addition, board drivers don't report data if the self-status indicates an error. The BMI088 is an exception because it doesn't have a good status to check, and valid GPS packets get sent including status in data packet.

PR #428, opened 2024-07-09
Ptt/fix set param float

avtoku:
set_param_float in param.cpp was sending the entire set of parameters in response to setting a single parameter instead of just sending back the updated value.


bsutherland333:
You had a commit related to your other pull request on this branch, but I fixed that. If everything still looks good to you @avtoku then go ahead and merge.

bsutherland333 (approved):

PR #429, opened 2024-07-10
Clang format changes

bsutherland333:
After reviewing the ROS2 and Google style guides, @avtoku and myself realized that the .clang-format file was not completely in line with their standards. This updates the .clang-format file and the code to fix that.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -11,7 +11,7 @@ AllowAllParametersOfDeclarationOnNextLine: false
 AllowShortBlocksOnASingleLine: Always
 AllowShortCaseLabelsOnASingleLine: false
 AllowShortFunctionsOnASingleLine: All
-AllowShortIfStatementsOnASingleLine: Always
+AllowShortIfStatementsOnASingleLine: AllIfsAndElse
 AllowShortLambdasOnASingleLine: All
 AllowShortLoopsOnASingleLine: true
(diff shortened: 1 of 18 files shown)


avtoku:
It looks like ROS2 has instructions for linting your code that go beyond a clang-format call.
Personally I'd be inclined to just leave it as the stock google .clang-format if we are not going to go all the way.

bsutherland333:
Ok, how about for now we just worry about the .clang-format file. Are you suggesting we leave it as is?

avtoku:
> Ok, how about for now we just worry about the .clang-format file. Are you suggesting we leave it as is?

For Code that is compiled and run in the ROS2 environment, ROS2 recommends using their ament linter: https://docs.ros.org/en/humble/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html#linters
This is beyond what I personally want to deal with for the rosflight_firmware.

For rosflight_firmware that lives outside the ROS2 environment we should not have to comply with ROS2 conventions. I think we should eventually use a few minor tweaks of the .clang-format Google format (like longer column limit), but for now let's just leave it as-is.

bsutherland333:
@avtoku How does that look?

avtoku:
> @avtoku How does that look?

Works for me.

avtoku (approved):
Looks good to me.

PR #434, opened 2024-08-27
Ptt/init reporting

avtoku:
Changes have been made to allow the board to send initialization status messages to the companion computer.
This requires also updating the board supports package in the varmint_h7 directory.

### DIFF
--- a/include/comm_manager.h
+++ b/include/comm_manager.h
@@ -162,7 +162,6 @@ private:
 
   // Debugging Utils
   void send_named_value_int(const char * const name, int32_t value);
-  //    void send_named_command_struct(const char *const name, control_t command_struct);
 
   void send_next_param(void);
 
@@ -182,7 +181,6 @@ public:
   void log(CommLinkInterface::LogSeverity severity, const char * fmt, ...);
   void log_message(CommLinkInterface::LogSeverity severity, char * text);
 
-//  void send_parameter_list();
   void send_named_value_float(const char * const name, float value);
 
   void send_backup_data(const StateManager::BackupData & backup_data);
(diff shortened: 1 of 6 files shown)


include/comm_manager.h
```diff
@@ -180,8 +180,9 @@ class CommManager : public CommLinkInterface::ListenerInterface, public ParamLis
   void send_param_value(uint16_t param_id);
   void update_status();
   void log(CommLinkInterface::LogSeverity severity, const char * fmt, ...);
+  void log_message(CommLinkInterface::LogSeverity severity, char * text);
 
-  void send_parameter_list();
+//  void send_parameter_list();
```
bsutherland333, line 185: It would be better to remove old code than comment it out

src/comm_manager.cpp
```diff
@@ -129,7 +129,10 @@ void CommManager::param_request_list_callback(uint8_t target_system)
   if (target_system == sysid_) { send_params_index_ = 0; }
 }
 
-void CommManager::send_parameter_list() { send_params_index_ = 0; }
+//void CommManager::send_parameter_list()
```
bsutherland333, line 132: Same as above

src/comm_manager.cpp
```diff
@@ -526,6 +534,23 @@ void CommManager::send_named_value_float(const char * const name, float value)
 
 void CommManager::send_next_param(void)
 {
+  //PTT Hack
```
bsutherland333, line 537: Should we figure out a more permanent solution for this before merging these changes into main?

bsutherland333 (changes requested):

include/comm_manager.h
avtoku: done

src/comm_manager.cpp
avtoku: done

src/comm_manager.cpp
avtoku: yes we should think of a more permanent solution

PR #435, opened 2024-09-25
Adjust baro data cal to just report raw sensor data

avtoku:
Barometers these days are very accurate, so bias offset calibration is now taken to be the pressure altitude at startup.
Raw baro pressure and 'raw' pressure altitude are reported in the telemetry packets.

### DIFF
--- a/src/sensors.cpp
+++ b/src/sensors.cpp
@@ -66,9 +66,6 @@ void Sensors::init()
 
   next_sensor_to_update_ = BAROMETER;
 
-  float alt = rf_.params_.get_param_float(PARAM_GROUND_LEVEL);
-  ground_pressure_ = 101325.0f * static_cast<float>(pow((1 - 2.25694e-5 * alt), 5.2553));
-
   int_start_us_ = rf_.board_.clock_micros();
 
   this->update_battery_monitor_multipliers();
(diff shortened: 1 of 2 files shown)


src/sensors.cpp
```diff
@@ -462,17 +459,19 @@ void Sensors::calibrate_baro()
...
         rf_.comm_manager_.log(CommLinkInterface::LogSeverity::LOG_ERROR,
-                              "Too much movement for barometer cal");
+                              "Too much movement for barometer ground pressure cal");
       }
       baro_calibration_mean_ = 0.0f;
       baro_calibration_var_ = 0.0f;
       baro_calibration_count_ = 0;
     } else if (baro_calibration_count_ > SENSOR_CAL_DELAY_CYCLES) {
-      float measurement = data_.baro_pressure - ground_pressure_;
+      float measurement = data_.baro_pressure;
```
iandareid, line 474: If we don't use the ground pressure, is there a reason to even run the ground pressure calibration? Or am I missing something?

avtoku: The bias (ground pressure) is computed to provide a ground altitude/pressure reference for cases where other references are not available. It is accessible to the companion computer by reading the PARAM_BARO_BIAS register. For convenience the standard atmosphere model is used to also compute the PARAM_GROUND_LEVEL parameter.

iandareid (approved):

PR #436, opened 2024-09-25
Changes pitot velocity to indicated air speed (ias)

avtoku:
Pitot velocity was trying to report the true air speed, but used the pitot sensor's internal temperature and the barometric pressure to compute the density. Unfortunately the pitot sensor temperature is not the atmospheric outside air temperature (OAT).
The computation is changed to compute the indicated air speed (IAS).

### DIFF
--- a/src/sensors.cpp
+++ b/src/sensors.cpp
@@ -561,10 +561,9 @@ void Sensors::correct_diff_pressure()
   if (!diff_pressure_calibrated_) { calibrate_diff_pressure(); }
   data_.diff_pressure -= rf_.params_.get_param_float(PARAM_DIFF_PRESS_BIAS);
 
-  float atm = 101325.0f;
-  if (data_.baro_present) { atm = data_.baro_pressure; }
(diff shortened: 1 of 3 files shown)


src/sensors.cpp
iandareid, line 1: @avtoku  Should we put a parameter for the air density 1.225? Just so if the density ever is of interest we don't have to reach in here and reflash.

avtoku: Ideally we'd just get rid of that field in the differential pressure packet but I didn't want to cause too many ripples. I would leave it as a constant 1.225 kg/m^3 since that does not change ever for computing the indicated air speed. If you want it to report true air speed, then I'd use the standard atmosphere model of density ratio as a function of pressure altitude (rather than a hand-coded value), and since we do not have a measure of outside air temperature. Having air density as a parameter would have bad consequences if you didn't know to set it and there was some incorrect value stored there.

iandareid (approved):

Issue #432, opened 2024-08-02
New mixing system


JMoore5353:
In the current system, the controller outputs (in Nm or N) are mixed using the mixing matrices defined in `mixer.h`. These matrices combine the output but do not take into account any of the propeller dynamics or parameters. 

This means that we are currently taking forces and moments and converting them directly to PWM values by saturation. A better approach might be to use the propeller dynamic equations from the textbook to convert the desired forces and moments to input voltages and then to a throttle setting.

JMoore5353:
@bsutherland333 @iandareid I have some thoughts I'd like your input on.

I implemented a mixing system that dynamically takes in the number of motors and the motor parameters to create the mixing matrix, using the Eigen library. It works well in sim. However, it dynamically allocates memory on the heap to do this (since the size of the matrices are defined at runtime). We can't use (or would strongly prefer not to use) a heap on the microcontroller, which makes this challenging.

A couple ideas:
- Precompute and hard code some mixing matrices for common frame types. 
    - Pros:
        - Doesn't use Eigen / dynamically allocated memory
        - Doesn't need to compute the mixing matrix on the microcontroller
    - Cons: 
        - Can't adapt for different motor positions (i.e., if we have  a big quad frame vs a small quad frame -- this would change the mixing matrix, but we wouldn't be able to update it unless we recompile).
        - Restricted to frames / configurations that we decide, needing a recompile for other configurations.
- Compute the mixing matrix in `rosflight_io` node and send each entry in the matrix as a parameter to the firmware
    - Pros: 
        - No dynamic memory allocation / extra computation in the firmware.
        - Completely flexible since we can do everything dynamically on the computer
        - Can take advantage of ROS2 array parameters to pass in motor locations (instead of passing them one by one to the firmware)
     - Cons: 
         - Have to add >=40 mixing matrix parameters (4 x `NUM_MIXER_OUTPUTS`)  to the firmware param list

I think my vote would be to do both of the above, meaning we have precomputed, hard coded matrices for specific configurations (like it is now), and then add a `custom_mixer` option that is loaded from params from `rosflight_io`.

What are your thoughts?

JMoore5353:
[...]

PR #438, opened 2024-10-14
432 new mixing system

JMoore5353:
Updated the mixer for multirotors to make it based on the Chapter 14 implementation. For multirotors, since the output of the controller is forces and torques, the mixer computes the desired angular speeds based on the motor model parameters. 

For fixedwing vehicles, the mixer architectre stayed the same, since ROSplane (and RC) commands come in as control surface delta commands, or as a throttle setting. 

Note that this PR does not expand the mixing definition to include forces in all three axes, and it doesn't add support for a dynamically-changing mixing matrix, both of which were mentioned in #432. The architecture currently only supports a "fixedwing" or "multirotor" mixing mode. The other changes will happen when we need to support them (i.e. for quadplane or tiltrotor).


include/mixer.h
bsutherland333, line 1: These mixing matrices use drastically different and far less intuitive numbers, could you add some explination as to where they came from and how to get them? Also, is it possible to normalize them back down to something along the lines of -1 - 1 and 0 - 1? I suppose I don't understand where they are coming from.

src/controller.cpp
```diff
@@ -38,6 +38,7 @@
 
 #include <cstdbool>
 #include <cstdint>
+#include <iostream>
```
bsutherland333, line 41: This probably needs to be removed

src/controller.cpp
```diff
@@ -244,7 +298,7 @@ float Controller::PID::run(float dt, float x, float x_c, bool update_integrator)
   if (dt > 0.0001f) {
     // calculate D term (use dirty derivative if we don't have access to a measurement of the
     // derivative) The dirty derivative is a sort of low-pass filtered version of the derivative.
-    //// (Include reference to Dr. Beard's notes here)
+    //// TODO: (Include reference to Dr. Beard's notes here)
```
bsutherland333, line 301: Should we add these references before merging?

src/controller.cpp
```diff
@@ -49,25 +50,58 @@ void Controller::init()
...
-  float max = RF_.params_.get_param_float(PARAM_MAX_COMMAND);
-  float min = -max;
+  // Calculate the max thrust to convert from throttle setting to thrust
+  calculate_max_thrust();
+
+  max_throttle_ = RF_.params_.get_param_float(PARAM_RC_MAX_THROTTLE);
+
+  // Don't saturate the torque values
+  float max_torque = INFINITY;
+  float min = -max_torque;
```
bsutherland333, line 60: Should we rename this to be min_torque, since max is being renamed to max_torque? That would make it more clear that they're related.

src/param.cpp
bsutherland333, line 1: How robust is the mixing equations to model mismatch? Although being able to get really good performance from a model based mixing is great, it would also be good if someone who didn't care as much could get a 'typical' multirotor flying without needing to adjust many of these parameters. At the very least, we should pick values that are both conservative and reasonable for typical multirotors as the default.

include/mixer.h
bsutherland333: I'm mostly just thinking about when someone wants to come through and modify the mixing matrix, a bunch of 0-1 values are pretty intuitive while I have no idea where these values came from.

bsutherland333 (dismissed):
Overall really great work, I think this will be a welcome improvement to ROSflight, especially once we have an easy way to configure the mixing from a base station.

iandareid:
I think that before we merge this we should look at ways to make the unit test not freak out at the build. I wonder if we could just make the compilation a little more lax? Give an exception for Eigen? I'm not sure.

include/mixer.h
avtoku, line 1: The PWM outputs are controlled by three hardware timers on the STM32 chip. Each timer controls multiple PWM's, but the frequency within each group must be the same.
On the Varmint channels 0-9 are grouped as (0, 1, 2, 3), (4, 5, 6, 7), (8, 9)
On Pixracer Pro channels 0-7 are grouped as (0, 1, 2, 3), (4, 5), (6, 7)

So, 
{490, 490, 490, 490,   490, 490,  490, 490,   50, 50} is OK for both Varmint and Pixracer, but
{490, 490, 490, 490,   490, 490,  50,   50,     50, 50}, is OK for Pixracer (which ignores the last two) but not for Varmint since the second group of 4 has mixed frequencies.
{490, 490, 490, 490,   490,   50,  50,   50,     50, 50}, does not work for either.

My suggestion is to arrange it so in all cases have the first 4 PWM's have the same frequency, and the second 4 have the same frequency in all mixing matrices.

Side note:
We could change the Varmint channels to be grouped as:
On the Varmint channels 0-9 are grouped as (0, 1, 2, 3), (8, 9), (4, 5, 6, 7)
so that all cases work for both Varmint and Pixracer, but all future revisions of Varmint will have channels 8&9 with isolated  power, ground, and signal, which is intended for powerful motor controllers that may have shifted ground signals or run off a separate power bus, and are annoying to use if you don't need that. If you need more isolated motor controller channels, there are motor controllers that have this isolation built in on the motor controller side.

avtoku:
I think the approach needs some more discussion.  I would rather start with just a mixer matrix that can be loaded from the companion computer, or from selected standard cases, and stored as parameters.

src/controller.cpp
JMoore5353: Yes it does. Good catch.

include/mixer.h
JMoore5353: That's a good point. These values are computed from the motor model parameters as in Section 14.4.3 in the Small Unmanned Aircraft book. They aren't intuitive values since the matrix transforms angular speeds into forces and torques (which isn't intuitive to me at least :smile:). I computed them with a Python script, though. Perhaps it would be useful to include the Python script for reference?

We could normalize them, but we'd need to multiply by the normalization factor out front for the math to work out properly. 

I changed this from the previous implementation for a couple reasons:
- In the previous approach, there wasn't a clear path from forces and torques -> desired angular speeds -> voltage setting -> throttle setpoint. Instead, the forces and torques were normalized and saturated a couple times so that they happened to be in between 0 and 1. This didn't make much sense to me, but it is possible I don't fully understand all the assumptions that went into the previous design. 
- This new approach is more in line with the approach in the book.

src/controller.cpp
JMoore5353: Those were notes from before... I don't know what they were referencing, so maybe we remove that line?

src/param.cpp
JMoore5353: I don't know the answer to that question.

include/mixer.h
JMoore5353: Sounds good. I can change that.

avtoku:
It looks like the equations for torque 14.7-14.9 have been simplified to assume the thrust vector of the individual rotors is along the z-axis. If a rotor is tilted off the z-axis, it can induce very large torques.

JMoore5353:
Adding the TODO items from our discussions here:

- [ ] Recompute the canned aircraft according to the new derivation.
- [ ] Add the ability to use the custom mixer based on motor and propeller equations
- [ ] Add computation of matrix on Jetson (either in `rosflight_io` or in the respective autonomy packages, `roscopter` and `rosplane`)
- [ ] Add functionality to switch between mixers based on if it is under RC or onboard control. 
- [ ] Add 60+ mixer parameters to the firmware (might add 2 sets, so we can have a custom safety mixer for RC control)
- [ ] Document the changes to the mixer

avtoku:
I limited the data size on the board to avoid large, erroneous, read/write requests hanging the main loop. We'll need to make sure to increase the storage allowance to accommodate the additional parameters.

include/command_manager.h
```diff
@@ -77,38 +86,50 @@ class CommandManager : public ParamListenerInterface
...
                            {false, ANGLE, 0.0},
                            {false, RATE, 0.0},
+                           {false, THROTTLE, 0.0},
+                           {false, THROTTLE, 0.0},
                            {false, THROTTLE, 0.0}};
   control_t offboard_command_ = {0,
                                  {false, ANGLE, 0.0},
                                  {false, ANGLE, 0.0},
                                  {false, RATE, 0.0},
+                                 {false, THROTTLE, 0.0},
```
iandareid, line 108: The extra throttles are just the forces commanded in x and y correct? If so is there a way to make this clear? Comment/name change? Or is it clear as you go to implement your own mixer?

src/comm_manager.cpp
```diff
@@ -237,36 +237,46 @@ void CommManager::offboard_control_callback(const CommLinkInterface::OffboardCon
...
-      new_offboard_command.x.type = PASSTHROUGH;
-      new_offboard_command.y.type = PASSTHROUGH;
-      new_offboard_command.z.type = PASSTHROUGH;
-      new_offboard_command.F.type = THROTTLE;
+      new_offboard_command.Qx.type = PASSTHROUGH;
+      new_offboard_command.Qy.type = PASSTHROUGH;
+      new_offboard_command.Qz.type = PASSTHROUGH;
+      new_offboard_command.Fx.type = PASSTHROUGH;
+      new_offboard_command.Fy.type = PASSTHROUGH;
+      new_offboard_command.Fz.type = PASSTHROUGH;
```
iandareid, line 263: Does the input from ROSplane need to change to accommodate this?

include/mixer.h
```diff
@@ -242,25 +267,41 @@ class Mixer : public ParamListenerInterface
...
+    &custom_mixing,
   };
 
   // clang-format on
 
+  // Define parameters for the mixer
+  float R_;                      // Motor resistance
+  float rho_;                    // Air density
+  float K_V_;                    // Motor back-emf constant
+  float K_Q_ = 0.01706;          // Motor torque constant
```
iandareid, line 281: Was this meant to be a hard-coded constant?

src/command_manager.cpp
```diff
@@ -103,17 +114,38 @@ void CommandManager::init_failsafe()
 void CommandManager::interpret_rc(void)
 {
   // get initial, unscaled RC values
-  rc_command_.x.value = RF_.rc_.stick(RC::STICK_X);
-  rc_command_.y.value = RF_.rc_.stick(RC::STICK_Y);
-  rc_command_.z.value = RF_.rc_.stick(RC::STICK_Z);
-  rc_command_.F.value = RF_.rc_.stick(RC::STICK_F);
+  // TODO: Adjust this to choose the channel that the RC thottle corresponds to
```
iandareid, line 117: Does this TODO need to be resolved or does the feature work and this will be dealt with later?

src/controller.cpp
```diff
@@ -49,25 +49,56 @@ void Controller::init()
...
+  float b = KV;
+  float c = i0 * R - V_max;
+
+  // Using Eq. 4.19 and setting equal to the equation at the beginning of 4.3 in Small Unmanned Aircraft
+  // We only need the positive root. Since a and b positive, sqrt term must be positive
+  float omega = (-b + sqrt(pow(b, 2.0) - 4 * a * c)) / (2 * a);
+
+  // Calculate the max thrust from Eq in 4.3 of Small Unmanned Aircraft
+  // Note that the equation is for a single motor, so to calculate max thrust, we need to multiply by the number of motors
+  max_thrust_ = rho * pow(D, 4.0) * CT * pow(omega, 2.0) / (4 * pow(M_PI, 2.0)) * num_motors;
```
iandareid, line 101: This is the good stuff.

src/controller.cpp
```diff
@@ -87,22 +118,27 @@ void Controller::run()
 
   // Check if integrators should be updated
   //! @todo better way to figure out if throttle is high
+  // TODO: fix this... Needs to be checked based on the throttle channel (not necessarily Fz)
```
iandareid, line 121: I assume this will be resolved by the time this is not a draft.

include/command_manager.h
iandareid: Coming back after seeing how this is used in the controller this makes more sense. I am still confused why they are all throttles wouldn't only one at a time be throttle? What am I missing?

src/controller.cpp
```diff
@@ -281,7 +371,6 @@ float Controller::PID::run(float dt, float x, float x_c, bool update_integrator,
   float u = p_term - d_term + i_term;
 
   // Integrator anti-windup
-  //// Include reference to Dr. Beard's notes here
   float u_sat = (u > max_) ? max_ : (u < min_) ? min_ : u;
   if (u != u_sat && fabs(i_term) > fabs(u - p_term + d_term) && ki_ > 0.0f) {
```
iandareid, line 375: Isn't this the integrator anti-wind up that caused us so much grief on the ROSplane controller?

src/mixer.cpp
```diff
@@ -34,23 +34,204 @@
 #include "rosflight.h"
 
 #include <cstdint>
+// We don't have a method to print the mixer currently being used for the user's benefit.
```
iandareid, line 37: Should this be resolved before merging? We could just add it to the TODOs for the project.

src/mixer.cpp
```diff
@@ -86,12 +393,253 @@ void Mixer::init_mixing()
...
+
 void Mixer::init_PWM()
 {
-  if (mixer_to_use_ != nullptr) {
-    RF_.board_.pwm_init_multi(mixer_to_use_->default_pwm_rate, NUM_MIXER_OUTPUTS);
+  if (mixer_to_use_.primary_mixer_ptr != nullptr) {
+    RF_.board_.pwm_init_multi(primary_mixer_.default_pwm_rate, NUM_MIXER_OUTPUTS);
   } else {
-    RF_.board_.pwm_init_multi(passthrough_mixing.default_pwm_rate, NUM_MIXER_OUTPUTS);
+    RF_.board_.pwm_init_multi(esc_calibration_mixing.default_pwm_rate, NUM_MIXER_OUTPUTS);
```
iandareid, line 642: Why the change here? Why a calibration instead of assuming passthrough?

src/mixer.cpp
```diff
@@ -132,55 +680,192 @@ void Mixer::set_new_aux_command(aux_command_t new_aux_command)
...
+                    commands.Qz * (*mixer_to_use_.Qz)[i];
+
+      // Save off the largest control output (for motors) if it is greater than 1.0 for future scaling
+      if ((*mixer_to_use_.output_type)[i] == M) {
+        if (abs(outputs_[i]) > max_output) { max_output = abs(outputs_[i]); }
+      }
+    }
   }
 
-  if (mixer_to_use_ == nullptr) { return; }
```
iandareid, line 151: Should we keep this check around?

include/command_manager.h
JMoore5353: That is a good question. They are all named all `THROTTLE` because they aren't rate or angle commands. These flags are used by the controller to determine if the commands should be multiplied by the `max_throttle` and/or the `max_thrust` parameters.

Since we don't know which one will be used by the controller (X, Y, or Z channel), I assigned them all to be throttle types.

When using the mixer, the constraint is either you use the onboard controller (with throttle settings for all X,Y,Z axes), or you use offboard controller that outputs forces directly (passthrough) to the mixer.

src/comm_manager.cpp
JMoore5353: No, other than renaming the command from "F" to "Fx". This just means it won't be scaled by the `max_thrust` and `max_throttle` parameters, which are really designed for multirotors anyway.

Also note that the fixedwing aircraft mixer doesn't take in forces and moments, really. It takes in delta commands, which correspond one to one with servo and motor setpoints. This means that the mixers for fixedwings are fundamentally different than for multirotors, since they don't need to be inverted, won't use motor parameters, etc. It's set up this way to avoid loading all the aerodynamic coefficients into the firmware.

include/mixer.h
JMoore5353: Nope. I don't know why that one escaped. I fixed that.

src/command_manager.cpp
JMoore5353: The feature works, but I forgot to remove the TODO

src/mixer.cpp
JMoore5353: I deleted passthrough mixing, since it is not used anywhere except for here, and wasn't actually passthrough (it was a bunch of zeros). Assigning it here is a placeholder until the user chooses the mixer they want (this branch is only run if we don't have a valid mixer, in which case, nothing else in the mixer gets run).

src/mixer.cpp
JMoore5353: No, it couldn't be called. The check lives at the top of the `mix_output` method, where it won't run the mixer if the mixer isn't selected. This check is still here, but not at line 705. I think GitHub jumbled the lines because I added a bunch of new ones.

JMoore5353:
@avtoku FYI I added the suggestions from our code review that I did not implement to the issues tab. These were things that I thought weren't as urgent for this PR. Thanks for the suggestions!

src/sensors.cpp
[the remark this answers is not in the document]
iandareid: But again I don't think that this should necessarily be part of this PR.

test/CMakeLists.txt
iandareid, line 1: Could we leave all of the flags except the one that is tripped by Eigen? That way we are doing a "minimal" change in a sense?

test/command_manager_test.cpp
```diff
@@ -85,14 +89,14 @@ TEST_F(CommandManagerTest, Default)
...
-  EXPECT_EQ(output.F.type, THROTTLE);
-  EXPECT_CLOSE(output.F.value, 0.0);
+  EXPECT_EQ(output.Qx.type, ANGLE);
+  EXPECT_CLOSE(output.Qx.value, 0.0);
+  EXPECT_EQ(output.Qy.type, ANGLE);
+  EXPECT_CLOSE(output.Qy.value, 0.0);
+  EXPECT_EQ(output.Qz.type, RATE);
+  EXPECT_CLOSE(output.Qz.value, 0.0);
+  EXPECT_EQ(output.Fz.type, THROTTLE);
+  EXPECT_CLOSE(output.Fz.value, 0.0);
```
iandareid, line 99: The Controller::Output makes this much more clear!

iandareid (commented):
Great and very understandable changes. I think we should talk about the flags more in depth. I think I may be missing something.

.github/workflows/release.yml
bsutherland333, line 1: If Eigen is included as a submodule the we don't need to install this package, right?

.github/workflows/pre-release.yml
bsutherland333, line 1: Same as other.

comms/mavlink/v1.0
bsutherland333, line 1: Is this still v1.0?

include/param.h
bsutherland333, line 1: Are we sure we don't want to implement an array type parameter so we don't need to have this many parameters?

bsutherland333 (approved):
This looks like a lot of really great work. I think this will be a great improvement and don't have much else to suggest, other than we should definitely spend some time testing this on real hardware and make sure every feature works as expected.

include/param.h
JMoore5353: I think we do. I opened a separate issue for that.

PR #446, opened 2025-03-22
Move sending of PWM from the mixer to the main loop. 

avtoku:
This edit makes it easier to see where PWM's are sent to the servos.


JMoore5353 (approved):
Looks good! I think there is a typo in the comment in mixer.h (const, not cont).

avtoku:
I fixed the typo. Note that I am not authorized to merge.

PR #452, opened 2025-06-12
Fixed state_.timestamp bug in estimator

JMoore5353:
This PR fixes a bug with the initialization of the `timestamp_us` field in the firmware estimator. This error was causing the ERROR_TIME_GOING_BACKWARDS error to trigger nonstop on boot.

Previously on startup (the first time through the estimator loop), the estimator initializes some variables and returns. The controller then queries some of those variables the first time through its loop. However, the `state_.timestamp_us` field was not initialized, so the controller was reading in random values for that variable.

I also updated the `prev_time_us_` variable in the controller. Otherwise, if the ERROR_TIME_GOING_BACKWARDS does get triggered, then the controller has no way to not trigger on the next iteration.

### DIFF
--- a/src/estimator.cpp
+++ b/src/estimator.cpp
@@ -91,7 +91,7 @@ void Estimator::reset_adaptive_bias()
 
 void Estimator::init()
 {
-  last_time_ = 0;
+  is_initialized_ = 0;
   last_acc_update_us_ = 0;
   last_extatt_update_us_ = 0;
   reset_state();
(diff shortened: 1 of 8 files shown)


src/estimator.cpp
avtoku, line 1: Doing the whole dt and time going backward logic in both Controller::run() and Estimator::run() is messy. I suggest moving that up into the if(got.imu) {...} in void ROSflight::run() with dt passed to Estimator::run( uint16_t dt). For the case of time going backward (you are in big trouble in this case), you just do not execute what is inside the if statement (or better yet trigger a failsafe mode?).

JMoore5353:
> This is a Band-Aid we should fix. Doing the whole dt and time going backward logic in both Controller::run() and Estimator::run() is messy. I suggest moving that up into the if(got.imu) {...} in void ROSflight::run() with dt passed to Controller::run() and Estimater::run()For the case of time going backward (you are in big trouble in this case), you just do not execute what is inside the if statement (or better yet trigger a failsafe mode?).

Good suggestion. I updated the

src/rosflight.cpp
avtoku, line 1: +Consioder adding a note by "if (got.imu && check_time_going_forwards())" that dt_ is computed buy check_time_going_forwards(). Otherwise, you wonder where it came from.

avtoku (approved):
Looks good, but the test board needs to be updated.

PR #457, opened 2025-08-04
Ptt/mixer mod

avtoku:
void Mixer::param_change_callback(uint16_t param_id) updates all the mixer values when only one needs to be updated. This pull request is intended to speed up the code and reduce unnecessary actions.


JMoore5353 (approved):
Looks good. I merged in #456, so I rebased onto main and fixed a typo in a var name.

PR #468, opened 2025-11-11
Use battery voltage as max voltage in mixer

JMoore5353:
This PR
1. Changes the mixer to use the currently measured voltage as the maximum voltage when converting mixer output (angular speeds) into desired motor commands.
2. Adds an LPF to the battery sensor. The parameters for the LPF were already present (`PARAM_BATTERY_VOLTAGE_ALPHA` and `PARAM_BATTERY_CURRENT_ALPHA`), and they were saved off as member variables, but were unused.

Previously, the mixer used a constant `VOLT_MAX` parameter to compute desired motor setpoint. Since the voltage drops over the course of the flight, this results in a rather large and increasing "disturbance" on the system, causing worse setpoint tracking as the flight continues.  

The LPF is added to reduce the noise in this mixer calculation.

### DIFF
--- a/src/sensors.cpp
+++ b/src/sensors.cpp
@@ -53,7 +53,7 @@ const float Sensors::DIFF_PRESSURE_MAX_CALIBRATION_VARIANCE = 100.0; // standard
 Sensors::Sensors(ROSflight & rosflight)
     : rf_(rosflight)
 {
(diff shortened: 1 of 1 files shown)


src/sensors.cpp
avtoku: It's unclear why the previous battery voltage is initialized to the voltage alpha.

PR #476, opened 2026-01-05
feat: add mag orientation params to firmware

### LINKED ISSUE #475: Pixracer pro mag orientation
The mag that we use for the Pixracer pro configuration is mounted to the M10034C GPS board, so it is separate from the Pixracer Pro. This means that the mag can be oriented in an arbitrary orientation from the IMU, which can mess things up. 

There is a rotation matrix hard-coded in the mag driver. However, this matrix should be used to rotate the mag measurements into the physical board frame, which still could be different than the vehicle frame. This is an issue for anything on the companion computer that expects the mag measurements to be in the body frame.

Therefore, I propose we add MAG_YAW, MAG_PITCH, and MAG_ROLL parameters to the firmware that configure how the magnetometer is oriented with respect to the vehicle. These parameters will not be used with the Varmint board, since the mag is hard mounted to the same board as the FC, but it is necessary for any other board configuration that has the mag on a board separate from the FC.

JMoore5353:
This PR adds three parameters to the firmware: MAG_ROLL, MAG_PITCH, and MAG_YAW. These are needed because some boards (e.g. Pixracer Pro) have mags that can be oriented at arbitrary angles with respect to the vehicle body. Runtime configuration of the mag orientation is desired in these cases.

The raw magnetic flux is rotated into the body frame according to the three parameters before being corrected by the hard and soft-iron calibration values.

Closes #475 .

Alternatively, this could have been accomplished by hard-coding the IST8308 rotation matrix and requiring users to point the mag in the body-frame X direction. This would enforce the MAG_ROLL,MAG_PITCH, and MAG_YAW to all be zero. This solution is a little more general, however.


avtoku:
Am I missing something here?  Prior to this, the IMU got rotated by PARAM_FC_ROLL|PITCH|YAW but the mag did not? That seems like it would cause all sorts of problems if you rotate an FCU with integrated MAG. This pull request lets the user fix that, but maybe changing PARAM_FC_ROLL|PITCH|YAW to PARAM_IMU_ROLL|PITCH|YAW is in order?

PR #487, opened 2026-04-23
Ptt/vscode

JMoore5353:
This PR (work done by @avtoku) adds CMake presets to improve the build process.

Also adds helpful files to help VScode users to get up and running quicker.


boards/varmint_h7/varmint_11X/specific/BoardConfig.h
```diff
@@ -40,8 +40,8 @@
 
 #include <CommonConfig.h>
 
-#define SANDBOX false
-#define BOARD_STATUS_PRINT false
+#define SANDBOX true
```
JMoore5353: @avtoku Did you mean to push sandbox mode? Same comment for the servo voltage below. Just checking.

JMoore5353:
Also, it looks like the scripts/run_tests.sh file does the same thing as the unit_tests.sh file in the root directory. Should we remove the unit_tests.sh file?

boards/varmint_h7/varmint_11X/specific/BoardConfig.h
avtoku: No, SANDBOX should be false.

avtoku:
> Also, it looks like the scripts/run_tests.sh file does the same thing as the unit_tests.sh file in the root directory. Should we remove the unit_tests.sh file?

unit_tests.sh in the root directory should be detelted.

boards/varmint_h7/varmint_11X/specific/BoardConfig.h
```diff
@@ -177,7 +177,7 @@ extern PCD_HandleTypeDef hpcd_USB_OTG_FS; // USB FS (48 MB/s)
 // Digital Potentiometer used in later versions
 //	#define MCP4017_I2C_ADDRESS 		(0x2F)
 #define MCP4017_I2C (&hi2c1)
-#define SERVO_VOLTAGE (4.8) // Volts
+#define SERVO_VOLTAGE (8.2) // Volts
```
JMoore5353: Does this var control what voltage the servos get? If so, we'll want to document that the Varmint 11X boards use 8 volt servos. I'm happy to document that if this is the case.

Also, since I opened this PR, I can't approve it, even though you were the one who did all the implementation. Sorry about that! This is my last question, and otherwise the PR looks good. If @avtoku you wouldn't mind approving it, then we can merge it in.

avtoku: Yes, it sets the voltage rail for the servos. I will set it back to 4.8V which is nominally safe and functional for most servos. We can add a parameter setting to control this as a TODO.

avtoku: Switching back to 4.8V

avtoku (approved):

PR #493, opened 2026-07-23
Update for Ubuntu 26.04 and ROS 2 Lyrical

mjhaskell:
The motivation for this pull request is to remove build warnings and errors when compiling on Ubuntu 26.04 for ROS 2 Lyrical.

Summary of Changes:

- Removed `cstdbool` header includes from C++ files. `cstdbool` was deprecated in C++ 17 (the build standard for Lyrical packages) and completely removed in C++ 20. C++ inherently includes the `bool` type and `true` and `false` keywords, so it is not necessary to include this header.
- Recursive include of `board.h` was removed from within `board.h`.
- CMake minimum policy version was bumped to the version currently set by `ros2 pkg create` with ROS 2 Lyrical (version 3.20).
- Included a CMake policy range for future proofing. Ubuntu 26.04 ships with CMake 4.2, but using the latest policies causes issues with GCC. Rather than trying to deal with debugging GCC and creating a pull request, I opted instead to back down the upper end of the policy range to 3.28 (which is what ships with Ubuntu 24.04). The version can be bumped to match the current version of CMake after the policy issues with GCC have been flushed out.
- Updated list of excluded directories in format script. Removed `Core`, `Drivers`, and `AL94_USB_Composite` since they no longer exist and added `boards` since this was Varmint-aided and it seems unsure how it should be formatted for now.
- ~~Ran `./scripts/fix-code-style.sh` to format code.~~ (removed after talking to Jacob so formatting can be done after other PRs are merged)

Note:

All tests passed when running the following test script:

``` bash
./scripts/run_tests.sh
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@
 # cmake --build build/test-release
 # ./build/test-release/test/unit_tests
 
-cmake_minimum_required(VERSION 3.8)
+cmake_minimum_required(VERSION 3.20...3.28)
 set(CMAKE_TRY_COMPILE_TARGET_TYPE "STATIC_LIBRARY")
 project(rosflight_firmware C CXX ASM)
 
--- a/boards/varmint_h7/common/CMakeLists.txt
+++ b/boards/varmint_h7/common/CMakeLists.txt
@@ -2,7 +2,7 @@
 
 ### project settings ###
 
-cmake_minimum_required(VERSION 3.8)
+cmake_minimum_required(VERSION 3.20...3.28)
 
 if(NOT DEFINED TARGET_NAME)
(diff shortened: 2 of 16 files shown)


mjhaskell:
Another note: the ROSflight documentation on the website has a specific order for including headers, but the code uses a different order (at least in some files - I did not check all files). One of them should likely be updated to the actual desired order.

mjhaskell:
@iandareid @JMoore5353

JMoore5353:
@mjhaskell Thanks for all this. I think we should fix the header order when we run the code style fix script.

mjhaskell:
@JMoore5353 I was thinking the same thing. I already opened a PR for formatting the code and left a comment their about the include order.

PR #494, opened 2026-07-24
Modernize cmake

mjhaskell:
This PR builds on #493. Review after reviewing #493.

The motivation of this PR is to modernize the CMake configuration of rosflight_firmware while #493 was to get a clean build.

Summary of Changes:

- Replace global CMake configuration with target-scoped properties
- Move rosflight firmware sources into a reusable library target
- Convert common STM32 H7 board setup into a helper function
- Add a scalable board registry for board selection
- Keep project setup limited to the top-level CMake file
- Normalize indentation and local variable naming
- Move Eigen includes from mixer header to mixer source file
- Rename boards/varmint_h7 to boards/stm32_h7

Note:

All tests passed when running the following test script:

``` bash
./scripts/run_tests.sh
```


mjhaskell:
@iandareid @JMoore5353

After talking to Jacob, I separated out formatting the code from the other PRs. I had to use some force pushes to get everything straight again.

mjhaskell:
@JMoore5353 Yes, the core firmware library was being defined here, in rosflight_sim, and in rosflight_io. It makes cmake easier to read when targets for a given project are completely defined in that location, so by creating libraries here it makes it much simpler to use elsewhere.

Also, I have since realized that rosflight_io uses the firmware very differently than rosflight_sim, so I actually have some more changes to push before I would recommend merging this PR. I am creating another library just for the generated mavlink headers, since that is all rosflight_io needs. I am also creating a library that solely contains the firmware version from the git commands so that all rosflight_io has to do is link to that library to have those variables defined.

mjhaskell:
I just pushed the other changes I previously mentioned to make the firmware work better with rosflight_io. rosflight_io only needed the generated mavlink headers as well as the compile definition of the firmware version, so I split those two things out into separate library targets.

Now rosflight_sim and rosflight_io do not have to repeat any firmware cmake configuration. In fact, the firmware is now quite portable where it could easily be removed as a submodule and changed to FetchContent or turned into a very lightweight ROS package (it wouldn't use anything from ROS - it would just install the firmware libraries so that sim and io could just call `find_package(rosflight_firmware)` then link to `rosflight_firmware::<library>`)...if that is a desired direction.

After talking to @iandareid, he wanted me to remove `#pragma GCC` commands from the source code where possible and move those to the cmake config.

@iandareid and I also discussed how there were two separate `mavlink.h` files that were a bit confusing. We thought it would be more clear to rename one of them. We did not want to touch generated code (content in `comms/mavlink/v1.0`), so we decided to rename `comms/mavlink/mavlink.h` to `mavlink_adapter.hpp` since it is actually firmware source code for adapting the generated mavlink headers to the rosflight firmware. Also, since it is firmware source code, we decided to move it into `include` and `src` rather than keep it in `comms/mavlink`. This had the additional benefit of providing a more consistent include path for mavlink files...before `comms/mavlink` and `comms/mavlink/v1.0` were both included paths. This meant all the generated headers were exposed from 2 locations...and they were included from both in different source files. Now the include paths are consistent: `#include "mavlink_adapter.hpp"` for the adapter and `#include "rosflight/mavlink.h"` for the generated header.

@JMoore5353 what do you think about these recent changes?

PR #495, opened 2026-07-27
Fetch Eigen with CMake and remove it as a Git submodule

mjhaskell:
This PR stacks on #494.

I separated this PR from #494 because I am unsure if there is a specific reason that Eigen is currently a submodule. Git submodules (as far as I understand) tend to indicate that the submodule is actively being developed, whereas CMake's FetchContent is intended for including a fixed version of a library that isn't being developed by this project.

Summary of Changes:

- Remove Eigen submodule
- Fetch Eigen with CMake


mjhaskell:
@iandareid @JMoore5353 Thoughts?

JMoore5353:
I'll ping @avtoku on this one. I seem to remember he preferred submodule over FetchContent at one point, but I can't remember exactly why (or if I'm misremembering :smile: ).

@mjhaskell Why are there so many Varmint file renames?

avtoku:
For those who use the STM32CubeIDE, it is (currently) not building with cmake, so that is why submodule might be preferred.

mjhaskell:
That makes sense. I separated this PR out from the others since it was mostly just a suggestion looking for feedback. I've been taught to generally prefer FetchContent, but if cmake isn't always used then I'd consider that a valid reason to use a submodule. Is that also why system Eigen isn't used in the firmware? (I don't know that I'd recommend system Eigen for the firmware simply because using the latest 5.0 version of Eigen removes SVD build warnings in the mixer, and Ubuntu 26.04 still ships with 3.4 - the question was mostly for my curiosity)

@JMoore5353 Those renames were not really part of this PR. They are from #498. I separated out #498 from #494, which required a force push to #494 and I forgot to update this PR with those changes. I can fix that so the renames are out of this PR, but if you guys want to stick with a submodule then I imagine this PR will just be closed.

mjhaskell:
To more easily support builds on Windows and using the STM32CubeIDE (without requiring CMake), we are leaving Eigen as a submodule.

- FetchContent only works with CMake
- System Eigen does not work well with Windows (inconsistent path to headers - not just `/usr/include/eigen3`)

PR #498, opened 2026-07-29
Rename boards for accuracy

mjhaskell:
This PR currently stacks on #494, but it could be altered to stack on main. The changes associated with this PR are really just the last 2 commits.

The current naming of the boards is not accurate (labeling them as "varmint") since it is only true for the 2 Varmint boards and not for the Pixracer Pro. This PR changes board names that are generic, which used Varmint, to use STM32 H7. This seems to be a more accurate description. Also, I think it sets a pattern for new boards to be added, e.g., there could be a `boards/stm32_f7` group.

I do not want to detract from the significant contributions made by the Varmint developers, so I tried to rename only what seemed to truly represent STM32 H7 boards.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@
 # cmake --build build/test-release
 # ./build/test-release/test/unit_tests
 
-cmake_minimum_required(VERSION 3.8)
+cmake_minimum_required(VERSION 3.20...3.28)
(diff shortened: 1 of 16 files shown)


mjhaskell:
@iandareid @JMoore5353

bsutherland333 (approved):
This is a logical change, just need to resolve current merge conflicts.

mjhaskell:
The build check seems to be frozen. I'm trying to figure out how to trigger it.

